PR #31, opened 2015-07-16
Fix syntax highlighting in README

DanielTomlinson:
GitHub doesn't deal well with root level indented code blocks in Markdown. (Treats the ```js as part of the code block)


pdubroy:
Thanks!

PR #32, opened 2015-07-16
visualizer: fix missing primitive characters

djdeath:
Just noticed some characters were missing in the top row of the lower view.


pdubroy:
Thanks for this! I just noticed it about an hour ago.

PR #48, opened 2015-12-09
Adding isOptional to Node, IteratorNode and Wrapper

mroeder:
Just as @alexwarth and I talked about last Friday, I added minMatches and maxMatches to the IterationNode (and its wrapper) so that semantic actions can act on this information (iteration can mean optional node or array of nodes).

I also extended the `_iter` test and documentation.

What I am not sure is whether the wrapper should use `minNumChildren` or `minNumMatches` (the IterationNode has a `minMatches` whereas the wrapper has `minNumChildren`).


doc/api-reference.md
```diff
@@ -204,6 +204,18 @@ Captures the portion of the input that was consumed by the node.
 
 The number of child nodes that the node has.
 
+<b><pre class="api">n.minNumChildren: number</pre></b>
+
+For an iterator node, the minimal number of child nodes possible.
```
pdubroy: I'd say "minimum" here, not "minimal".

doc/api-reference.md
```diff
@@ -204,6 +204,18 @@ Captures the portion of the input that was consumed by the node.
...
+
+<b><pre class="api">n.maxNumChildren: number</pre></b>
+
+For an iterator node, the maximal number of child nodes possible.
```
pdubroy: Likewise, "maximum".

doc/api-reference.md
```diff
@@ -204,6 +204,18 @@ Captures the portion of the input that was consumed by the node.
...
+
+<b><pre class="api">n.primitiveValue: number|string|...</pre></b>
+
+For a terminal node, the primitive value that the node has.
```
pdubroy: Maybe "the raw value that was consumed from the input stream"?

src/Semantics.js
```diff
@@ -96,6 +96,30 @@ Object.defineProperties(Wrapper.prototype, {
...
   numChildren: {get: function() { return this._node.numChildren(); }},
 
+  // Returns the minimum number of children for this CST node, if it's an iterator node.
+  // Otherwise, throws an exception.
+  minNumChildren: {
+    get: function() {
+      if (this.isIteration()) {
+        return this._node.minMatches;
+      }
+      throw new TypeError(
```
pdubroy: Is there any scenario where it might make more sense to return 0 here instead of throwing an exception? I can't think of one, just a thought.

test/test-semantics.js
```diff
@@ -264,6 +264,14 @@ test('_iter nodes', function(t) {
   });
   t.equal(s(m).op(), 'a,b,c');
 
+  s = g.semantics().addOperation('op', {
+    letters: function(ls) {
+      return [ls.minNumChildren, ls.maxNumChildren, ls.numChildren];
+    }
+  });
+  t.deepEquals(s(m).op(), [0, Number.POSITIVE_INFINITY, 3],
```
pdubroy: Maybe add a couple more test cases for other kinds of iter node?

doc/api-reference.md
mroeder: Well, I spent some time investigating whether minimum or minimal would be right here ;-) but I am happy to change that knowing that a native speaker corrected it.

src/Semantics.js
mroeder: The problem is that `minNumChildren` doesn't make sense for anything else than iterator nodes. Any other node has an exact number of children (and an iteration node for optional matches).

pdubroy: I guess it depends on what you intend this would be used for. Depending on that, I could imagine that, e.g. an Apply node would return 0 and INFINITY for min and max, or just return min == num == max. It would mean you could just call these methods without having to do a type check first.  But if you don't think it makes, don't worry about it.

Can you actually give an example of a use case for these methods? Is it just another way to check if the node comes from a `*`, `+`, or `?`, or is it more generally useful?

mroeder: Mhmm...

The whole discussion started with me being surprised that `ClassDecl = class className (extends className)? (with Formals)? ";"` leads to a semantic action that has 7 instead of 5 arguments (which I see when I look at it). Wanting to know whether something is optional (0-1x) or a repetition (0+ or 1+), I was looking at the grammar's rule body for `ClassDecl` where I only found 5 parts. Matching those 5 to 7 is hard(er) later on so Alex and I were thinking that putting `min` and `max` into the nodes/wrappers. We decided against having the iterator reporting the `operator` token (`?`, `*` or `+`) because future versions of Ohm could have `{2,5}` just like RegExps do.

Thinking about it now, it might make more sense to have `isOptional` instead of `minNumChildren` which will also make sense for any other kind of node (NO, not optional). Yet it provides less information.

Thoughts? @pdubroy @alexwarth (I actually favor `isOptional` now.)

pdubroy:
`isOptional` also seems ok, but I'd find it easier to think about in the context of a specific example of a semantic action that you'd like to write. Inside an action, you've got a value, and I'm having trouble coming up with an example of where it would be useful to know if that node was optional or not.

mroeder:
The only example I can currently come up with is the AST generator that I am currently working on.
Here an _optional_ node/attribute often means `null` or `value` and not `[]` or `[value]` compared to a proper repetition that should be a list but might as well be an empty list (`[]`).

Example code:
```javascript
{
  '_iter': function(children) {
    if ((this.minNumChildren === 0) && (this.maxNumChildren === 1)) {
      if (this.numChildren === 0)
        return null;
      else
        return children[0].toAST(this.args.mapping);
    }

    return children.map(function(child) {
      return child.toAST(this.args.mapping);
    }, this);
  }
}
```

pdubroy:
That seems like a reasonable use case -- I think it would make sense to add `isOptional`. Before 1.0 though, I want to think about re-organizing some of these helper methods on nodes, because right now they potentially clash with attribute/operation names.

alexwarth:
I like to think of them as built-in attributes and operations. So clashing
may not be the problem, but we should make sure that they behave the same
way as user defined operations/attributes. Let's talk about this on Monday!

mroeder:
@pdubroy @alexwarth Please review these changes to `isOptional`. If okay, I will do the merge.

pdubroy:
Would you mind squashing the commits into a single one? I find it hard to review the net effect of the changes this way. Plus, there's no need to have the commits that you ended up reverting locally still show up in the commit graph on master.

doc/api-reference.md
```diff
@@ -204,6 +204,14 @@ Captures the portion of the input that was consumed by the node.
 
 The number of child nodes that the node has.
 
+<b><pre class="api">n.isOptional() &rarr; boolean</pre></b>
+
+`true` if the node is an iterator node and can have no or one children (? operator), otherwise `false`.
```
pdubroy: I'd say "`true` if the node is an iterator node having either one children or no children"

src/Semantics.js
```diff
@@ -83,6 +83,13 @@ Wrapper.prototype.isLexical = function() {
   return this.isNonterminal() && this._node.isLexical();
 };
 
+// Returns `true` if the CST node associated with this wrapper is an iterator node
```
pdubroy: This comment is inconsistent with the documentation.

pdubroy:
LGTM if you fix up the comments to be consistent.

alexwarth:
LGTM too.

mroeder:
Merge by e24a146b2c22a719e48f0c84426ab6a69c630338

PR #50, opened 2016-01-15
Refactoring ES5 grammar to be more idiomatic

mroeder:
The original ES5 grammar for Ohm is based on OMeta's grammar which in fact is very close to the Appendix A of the ECMAScript 5.1 spec (ECMA-262). This makes it less idiomatic to Ohm and a manual creation of this grammar would look very different, I think.
Here are some refactorings to make the ES5 grammar use more of the Ohm features (not available in grammar notation of the spec).

### DIFF
--- a/examples/ecmascript/es5.ohm
+++ b/examples/ecmascript/es5.ohm
@@ -273,15 +273,13 @@ ES5 {
                     | "(" Expression ")"  -- parenExpr
 
   // @returns an ArrayExpr AST
-  ArrayLiteral = "[" AssignmentExpressionOrElisionComma* "]"                       -- trailingComma
-               | "[" AssignmentExpressionOrElisionComma* AssignmentExpression "]"  -- noTrailingComma
-  AssignmentExpressionOrElisionComma = (AssignmentExpression | Elision) ","
-  Elision =
+  ArrayLiteral = "[" ListOf<AssignmentExpressionOrElision, ","> "]"
+  AssignmentExpressionOrElision = AssignmentExpression
+                                |                       -- elision
 
   // @returns an ObjectExpr AST
-  ObjectLiteral = "{" PropertyAssignmentComma* "}" -- trailingComma
-                | "{" PropertyAssignmentComma* PropertyAssignment "}" -- noTrailingComma
-  PropertyAssignmentComma = PropertyAssignment ","
+  ObjectLiteral = "{" ListOf<PropertyAssignment, ","> "}"              -- noTrailingComma
+                | "{" NonemptyListOf<PropertyAssignment, ","> "," "}"  -- trailingComma
 
   // @returns a *Prop AST
   PropertyAssignment = get PropertyName "(" ")" "{" FunctionBody "}"                  -- getter
@@ -311,14 +309,7 @@ ES5 {
                  | MemberExpression Arguments         -- memberExpExp
 
   // @returns an array of ASTs
-  Arguments = "(" ")"                -- empty
-            | "(" ArgumentList ")"   -- nonEmpty
-
-  // @returns an array of ASTs
-  ArgumentList = AssignmentExpressionComma* AssignmentExpression -- many
-               | AssignmentExpression                            -- one
-
-  AssignmentExpressionComma = AssignmentExpression ","
+  Arguments = "(" ListOf<AssignmentExpression, ","> ")"
 
   LeftHandSideExpression = CallExpression
                          | NewExpression
@@ -432,17 +423,10 @@ ES5 {
   VariableStatement = var VariableDeclarationList #(sc)
 
   // @returns an array of ASTs
-  VariableDeclarationList = VariableDeclarationComma* VariableDeclaration -- many
-                          | VariableDeclaration
-
-  VariableDeclarationComma = VariableDeclaration ","
+  VariableDeclarationList = NonemptyListOf<VariableDeclaration, ",">
 
   // @returns an AST
-  VariableDeclaration = InitPattern  -- initialValue
-                      | identifier   -- noInitialValue
-
-  // @returns an AST
-  InitPattern = identifier Initialiser
+  VariableDeclaration = identifier Initialiser?
 
   // @returns an AST
   Initialiser = "=" AssignmentExpression
@@ -454,8 +438,7 @@ ES5 {
   ExpressionStatement = ~("{" | function) Expression #(sc)
 
   // @returns an AST
-  IfStatement = if "(" Expression ")" Statement else Statement  -- ifThenElse
-              | if "(" Expression ")" Statement                 -- ifThen
+  IfStatement = if "(" Expression ")" Statement (else Statement)?
 
   // @returns an AST
   IterationStatement = do Statement while "(" Expression ")" #(sc)  -- doWhile
@@ -515,10 +498,7 @@ ES5 {
     | function "(" FormalParameterList ")" "{" FunctionBody "}"             -- anonymous
 
   // @returns an array of ASTs
-  FormalParameterList = FormalParameter (CommaFormalParameter)* -- many
-                      |                                         -- zero
-
-  CommaFormalParameter = "," FormalParameter
+  FormalParameterList = ListOf<FormalParameter, ",">
 
   // @returns an AST
   FormalParameter = identifier


examples/ecmascript/es5.ohm
```diff
@@ -273,15 +273,12 @@ ES5 {
...
-  Elision =
+  ArrayLiteral = "[" ListOf<AssignmentExpressionOrElision, ","> "]"
+  AssignmentExpressionOrElision = AssignmentExpression
+                                |                       -- elision
 
   // @returns an ObjectExpr AST
-  ObjectLiteral = "{" PropertyAssignmentComma* "}" -- trailingComma
-                | "{" PropertyAssignmentComma* PropertyAssignment "}" -- noTrailingComma
-  PropertyAssignmentComma = PropertyAssignment ","
+  ObjectLiteral = "{" ListOf<PropertyAssignment, ","> ","? "}"
```
alexwarth: This doesn't have the same meaning as the original b/c it would happily accept `{,}` as an object literal. Here's a version that does:

```
ObjectLiteral
  = "{" ListOf<PropertyAssignment, ","> "}"  -- noTrailingComma
  | "{" NonemptyListOf<PropertyAssignment, ","> "," "}"  -- trailingComma
```

Unfortunately it's not as beautiful, but that's JavaScript for you :/

examples/ecmascript/es5.ohm
```diff
@@ -338,18 +327,13 @@ ES5 {
                   | "!"    UnaryExpression  -- lnot
                   | PostfixExpression
 
-  MultiplicativeExpression = MultiplicativeExpression "*" UnaryExpression -- mul
-                           | MultiplicativeExpression "/" UnaryExpression -- div
-                           | MultiplicativeExpression "%" UnaryExpression -- mod
+  MultiplicativeExpression = MultiplicativeExpression ("*"|"/"|"%") UnaryExpression -- mul
```
alexwarth: IMO the original version gives you a better API for writing semantic actions. (With this version, you might have to write a `switch` or chain of `if`s to figure out what to do, etc.)

examples/ecmascript/es5.ohm
```diff
@@ -273,15 +273,12 @@ ES5 {
                     | "(" Expression ")"  -- parenExpr
 
   // @returns an ArrayExpr AST
-  ArrayLiteral = "[" AssignmentExpressionOrElisionComma* "]"                       -- trailingComma
-               | "[" AssignmentExpressionOrElisionComma* AssignmentExpression "]"  -- noTrailingComma
-  AssignmentExpressionOrElisionComma = (AssignmentExpression | Elision) ","
-  Elision =
+  ArrayLiteral = "[" ListOf<AssignmentExpressionOrElision, ","> "]"
+  AssignmentExpressionOrElision = AssignmentExpression
+                                |                       -- elision
```
alexwarth: See my comment for `ObjectLiteral` -- this has the same problem.

examples/ecmascript/es5.ohm
```diff
@@ -311,20 +308,12 @@ ES5 {
...
-
-  AssignmentExpressionComma = AssignmentExpression ","
+  Arguments = "(" ListOf<AssignmentExpression, ","> ")"
 
   LeftHandSideExpression = CallExpression
                          | NewExpression
 
-  PostfixExpression = LeftHandSideExpression #(spacesNoNL "++")  -- postIncrement
-                    | LeftHandSideExpression #(spacesNoNL "--")  -- postDecrement
+  PostfixExpression = LeftHandSideExpression #(spacesNoNL ("++"|"--"))  -- post
```
pdubroy: I think it's also better to keep these cases separate, for convenience in writing semantic actions.

pdubroy:
Most of the ListOf changes look good to me. For the changes that remove inline rules in favour of a single rule containing a big Alt -- I'd be tempted to leave those as-is, for the reason Alex suggested. Also, the official ES5 grammar (https://es5.github.io/A.html#A.1) also presents these as separate cases, and I think it's nice to keep a 1:1 correspondence with the official grammar where possible. (But I think using ListOf instead of the recursive productions used in the official grammar is nicer.)

alexwarth:
Right, the only thing to watch out for are these cases where we would allow
a single comma (or whatever separator is used in the original ES6 grammar
rule) as a valid empty list.

Thanks for doing this, Marko!

examples/ecmascript/es5.ohm
mroeder: Point taken ;-).

examples/ecmascript/es5.ohm
mroeder: In this case it's spec compliant: You can have as many `,` in `[]` as you like.
`[]` = empty array, length 0
`[,]` = empty array, length 1
`[,,]` = empty array, length 2, etc.

alexwarth: Ok, cool. I didn't know that ;)

mroeder:
Mhmm... I disagree with the big alt because I have implemented a JS interpreter and transpiler before and the single big alt seems much more convenient (yes, it's a `switch` or `if`).
However, if I am really overruled here, I will revert those changes gnawingly :-p.

And by the idea of whole request: Keeping a 1:1 correspondence with the Appendix A and B leads to this rather ugly grammar. :-(

examples/ecmascript/es5.ohm
pdubroy: WAT

mroeder:
```javascript
[,,,,,,,,,,,,].join( 'hero'-1) + "Batman";
```

examples/ecmascript/es5.ohm
```diff
@@ -338,18 +328,13 @@ ES5 {
...
-  MultiplicativeExpression = MultiplicativeExpression "*" UnaryExpression -- mul
-                           | MultiplicativeExpression "/" UnaryExpression -- div
-                           | MultiplicativeExpression "%" UnaryExpression -- mod
+  MultiplicativeExpression = MultiplicativeExpression ("*"|"/"|"%") UnaryExpression -- mul
                            | UnaryExpression
 
-  AdditiveExpression = AdditiveExpression "+" MultiplicativeExpression -- add
-		     | AdditiveExpression "-" MultiplicativeExpression -- sub
-		     | MultiplicativeExpression
+  AdditiveExpression = AdditiveExpression ("+"|"-") MultiplicativeExpression -- add
```
pdubroy: With cases like this, I've never known what to use for the inline rule name. `AdditiveExpression_add` just seems weird. But it just occurred to me -- what about "rec"?

alexwarth: That's what I used to call these inline rules, but I was never super happy
with it. How about we just go with "rec" and make a mental note to revisit
this naming convention?

mroeder: We only have to name it `rec` if I get through with my "shorten all the alternatives to a big alternation" :-p.

pdubroy:
Re: "big alt" -- for cases that are truly uniform and the only difference is the operator (e.g. `++`/`--`), I'm fine with making them part of the same rule. For things that are semantically very different, but just happen to appear in the same part in the grammar, like `instanceof` and `in`, I think they should remain separate cases.

One advantage of the separate actions is that you have an up-front check that you have handled all the cases -- which you don't get with a switch statement. The big advantage of the switch statement that I see is that you can more easily share code between the cases.

mroeder:
Re: "big alt" -- I absolutely agree with that. That's why I originally did not change the UnaryExpression.
Sorry though that I got the `instanceof` and `in` mixed into the _RelationalExpression_.

alexwarth:
Hey guys! What's the latest on this refactoring? Are we stalled?

mroeder:
You can get an overview at: https://github.com/cdglabs/ohm/pull/50/files ;-). Still my suggestion and the last I heard we were indecisive about squashing all the similar cases into one (with a possible `switch` in their semantics).

I am worried that those changes generally will break some Ohm users' implementations and therefore I would first extract the grammar as `ohm-grammar-ecmascript`, have the Ohm repository depend on it and make the first version of it the current file. Then integrate this issue's changes as a second minor revision.

pdubroy:
I would not worry about it too much if people are depending on a broken grammar in a not-yet-1.0 project.

Also I already published it as a separate package on NPM but decided not to break it out into its own repo yet as I think it's the best "real" example we have.

mroeder:
The best real example could still mean that we rely on it - in examples - and but externalize the grammar and add it to the dev dependencies.

pdubroy:
If you'd really like to do that, go ahead. I don't think it should block this PR though.

pdubroy:
So we need to decide what to do about squashing the different cases into one (e.g., https://github.com/cdglabs/ohm/pull/50/files#diff-65095c55cdc04ca64238197a5263ece2R331). I'm still slightly in favour of leaving it the way it was before -- that's what we do in the other examples, and I don't see a compelling reason to switch. @alexwarth, do you want to make the final call?

alexwarth:
I also think that it would be better to "unsquash" these rules.

mroeder:
Reverted the squashing *groan* and checked against https://github.com/michaelficarra/everything.js.
Ready to merge now, @pdubroy @alexwarth?

alexwarth:
Please do!

PR #51, opened 2016-01-15
Introducing Ohm extras - useful extensions to the Ohm core

mroeder:
Not to clutter the Semantic and MatchResult objects with useful operations that are not always needed, the _Ohm extras_ will be a good place to put extras.
Here is a configurable CST to AST operation - `toAST(mapping)` - that can create e.g. ESTree-compatible ASTs for JavaScript. (https://github.com/estree/estree).


mroeder:
Yes at:
* `tape test/**/*.js`
* `var extras = require('ohm-js/extras');` :+1:
* documentation => `doc/extras/toAST.md` (with three different folders, would it make sense to put tests, docs and the actual code all under `extras`?)

alexwarth:
>> with three different folders, would it make sense to put tests, docs and the actual code all under `extras`?

I don't have a strong preference here, up to you guys :)

extras/index.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+module.exports = {
```
pdubroy: Test (ignore this comment)

mroeder:
> I'd give this a more descriptive name to indicate that it is returning a new semantics instance. Alternatively, you could just make it a function that creates a new semantics and immediately invokes the operation -- this might be the more common use case in practice.

Playing around with it for a while now, I have had case where I needed to do special rules for certain cases and I did this by first creating the `toAST()` operation and then extending the semantic:
```javascript
var s = ohm.extras.toAST(g);
s = g.extendSemantics(s).extendOperation('toAST', {
  SpecialCase: function(child1, child2, ...) { ... }
});
var res = g.match(...);
s(res).toAST({ /* mapping */ });
```

That said, I am not very happy with this solution. I experimented with a second argument:
```javscript
ohm.extras.toAST(g, /* extendOperation content goes here */);
```
and I could also push it down to the mapping at some point:
```javascript
s(res).toAST({
  // regular mapping
  SpecialCase: function(child1, child2, ...) { ... },
  // more regular mapping
});
```
I am slightly in favor of the latter one and it could lead to a direct call to `ohm.extras.toAST(res /* carries g in res._cst.grammar */, mapping)` though it does not make use of Ohm's general extension mechanisms. Thoughts?

pdubroy:
Both the 2nd one and the last one seem reasonable to me.

mroeder:
> I'd be tempted to keep this as a separate module for now -- i.e., make it so that you do var extras = require('ohm-js/extras'); to use it. What do you think?

Downside: browserify does not automatically bundle the extras into `dist/ohm.js` and `dist/ohm.min.js` anymore. (just tried that out but I think there may be a flag for that?!?)

pdubroy:
Eventually I would imagine that we would distribute the extras as a separate bundle.

extras/semantics-toAST.js
```diff
@@ -0,0 +1,108 @@
...
+// --------------------------------------------------------------------
+
+var pexprs = require('../src/pexprs');
+var MatchResult = require('../src/MatchResult');
+
+// --------------------------------------------------------------------
+// Operations
+// --------------------------------------------------------------------
+
+function toAST(res, mapping) {
```
pdubroy: It would still be nice to have at least a couple of sentences of documentation here, to describe what this function does and how it works. I still don't fully understand exactly how the mapping works.

mroeder:
@alexwarth @pdubroy Last chance for comments! ;-)
Now also on the documentation.

pdubroy:
I'd still like a bit of doc in the code :-) Just a few lines explaining what toAST is and what the arguments are is fine. Feel free to point to the user doc for more details...but when you encounter an undocumented function in the source code, you shouldn't have to go hunting through the user documentation in hopes of figuring out how it works.

pdubroy:
Otherwise LGTM.

mroeder:
> > I'd be tempted to keep this as a separate module for now -- i.e., make it so that you do var extras = require('ohm-js/extras'); to use it. What do you think?

> Downside: browserify does not automatically bundle the extras into dist/ohm.js and dist/ohm.min.js anymore. (just tried that out but I think there may be a flag for that?!?)

After fixing that, I looked around for quite a while now and there does not seem to be a way to do something similar to `require('ohm-js/extras')` in the browserified `dist\ohm(.min).js`. For that reason, I would like to put it back into `ohm.extras`. Counter-arguments? @alexwarth @pdubroy

pdubroy:
I would think they should go in a separate file. If they are truly extras, they probably shouldn't be bundled in the core script.

mroeder:
I invested some serious time here yesterday and the only thing I could find is https://github.com/substack/browserify-handbook#partitioning which would leave us with three files: `ohm-core.js`, `ohm.js` and `ohm-extras.js`. This seems unfortunate and you would have to include `ohm-core.js` before using any of the other files.
In my opinion - and that is the case for `toAST` - extras are still closer to Ohm than grammars, etc. and therefore use core classes and concepts that are otherwise "hidden". That makes it hard to partition those the basic ohm library from the extras.

pdubroy:
Yes, it seems like it would be definitely tricky when you are depending on
"internal" modules. In general I would expect the extras to be more like
the visualizer -- "close" to Ohm but not depending on internal modules. It
looks like all of the things you are importing are not really required,
just being used for type checks. If you eliminate those (e.g. just check
that the object has the methods you need) then it should be straightforward
to compile it into a separate bundle.

mroeder:
Or I add it to the public interface to Ohm... :-p

alexwarth:
Hey guys! How about we talk about this over Skype next week, and come up
with a plan together?

PR #60, opened 2016-02-14
Don't duplicate formal arguments

### LINKED ISSUE #54: Eliminate multiple underscore arguments in tests and examples
From the mdn page on [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode)

> strict mode requires that function parameter names be unique

The house style so far uses underscores for every argument that is not actually used in a semantic action, but this throws errors when run in strict mode. It might be better style to use `function (_1, _2, _3) {}` (or some other alternative?) instead.

jwmerrill:
Instead, give unused arguments distinct, descriptive names that start with a leading underscore. This pattern is preferable because it is compatible with strict mode.

Fixes #54.

Note: I checked that the math and viz examples continued to work after this change. The pl0 example was broken before I started; I think it needs maintenance to be brought up to speed with master, but I'm probably the wrong person to do it.

### DIFF
--- a/examples/math/index.html
+++ b/examples/math/index.html
@@ -268,15 +268,15 @@ s.addOperation(
       /*
         The following semantic actions are more of the same...
       */
-      MulExp:        function(e)       { return e.interpret(); },
-      MulExp_times:  function(x, _, y) { return x.interpret() * y.interpret(); },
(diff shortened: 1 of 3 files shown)


pdubroy:
Thanks!

PR #61, opened 2016-02-14
Add 'use strict' annotation to test-ohm-syntax

### LINKED ISSUE #54: Eliminate multiple underscore arguments in tests and examples

jwmerrill:
And fix duplicate _ arguments.

Fixes #54.

### DIFF
--- a/test/test-ohm-syntax.js
+++ b/test/test-ohm-syntax.js
@@ -1115,7 +1115,7 @@ test('inheritance', function(t) {
               'G2 <: G1 { foo = "bar" }'
             ]);
           },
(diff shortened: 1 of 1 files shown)


alexwarth:
Thanks!

PR #69, opened 2016-02-24
Include grammar source in recipes if possible.

pdubroy:
@mroeder and @alexwarth -- please take a look. This commit ensures that the grammar source is retained even when the grammar comes from a recipe. The impetus for this is to be able to show the definition for built-in rules inside the editor. I'm currently relying on `PExpr.prototype.toString()`, which gives a poor user experience.


alexwarth:
Hi Pat,

Retaining the grammar source in the recipe is a great idea! I've taken a
quick look at the code, and plan on taking a more careful look after my
class today. In the meantime, I have a question for you: why do we need
PExpr.prototype.toString() at all? Can't we just use the original source of
the grammar?

Cheers,
Alex

pdubroy:
With what I've implemented so far, only the top level expression of a rule
body has the source attached to it. If I made it so that every
subexpression also had an interval, I think you're right that we could get
rid of PExpr.prototype.toString(). It would make the recipes larger, but
I'm not sure that matters all that much.

alexwarth:
> If I made it so that every
subexpression also had an interval, I think you're right that we could get
rid of PExpr.prototype.toString(). It would make the recipes larger, but
I'm not sure that matters all that much.

Yeah, I think we should do that. I don't think the size of a recipe is a big deal, and as you said before, using `PExpr.prototype.toString()` is bad for the user experience.

mroeder:
Hey Pat and Alex,
Seeing this I am not sure what the intended use of the recipes is. So far, I thought that recipes are a different representation of on Ohm grammar that is actually closer to JS and can therefore be instantiated faster. From a "data perspective", putting in the grammar twice - as text and "JS recipe" - seems a bad idea to me since both could easily get out of since if handled manually.

However, I can see your interested in this "mapping" and why tools would want to know "where" to find a token when it's coming from a recipe.

So my question is: What is the intended use for recipes? (Should recipes be documented somewhere?)
From there I think we could see what is a good technique to get this mapping, maybe without duplicating information (and I am also not worried about the size of a recipe here).

alexwarth:
Hi Marko,

History time!

I came up with the idea for recipes to support the bootstrapping process. The Ohm grammar is written in Ohm, so we needed a way to load that into the system before we could do anything with other Ohm grammars.

Once we had recipes, I thought *hey, why not let people convert their own grammars to a recipes*? Instantiating an Ohm grammar from its recipe is significantly faster than parsing that grammar and invoking the appropriate semantic action to instantiate the grammar object. So it seemed like a good idea for uses where people care about loading time, especially when the grammar is "large", like ES5. Not sure how important this is to Ohm users, but that's what I was thinking.

> Should recipes be documented somewhere?

If the answer to my previous question is *yes*, then yes :)

The data duplication issue doesn't bother me very much. It would if there were any chance that the two could get out of step, but I'm sure that our tools can guarantee that it will never happen. (No one should be editing recipes manually anyway.) And while I'm sure we can come up with more efficient ways to keep track of the mapping from rules / terms / etc. to source text, at this point I'd rather do the easiest thing that will support the PX that we want, and revisit this later if it becomes an issue. Bigger fish to fry, etc. Though sometimes the smaller fish are tastier. But I digress.

Cheers,
Alex

mroeder:
Thanks again, Alex!
So all is as I imagined only that the recipes are not seen as true _alternatives_ to the text grammars (that's what I am reading out of your response). The need for bootstrapping was clear to me before and I think it's a neat way of doing that.

I am troubled by the duplication ;-) and therefore reserve the write to quote you:
> No one should be editing recipes manually anyway.

There was a quote in history that sounded similar... \*foraging*:
> No one will need more than 637 kB of memory for a personal computer.

If people can do it, they will!! :-)

pdubroy:
@alexwarth: Thanks for the suggestion, I will fix this so that all expressions are linked with the source.

@mroeder: Some good points. I view the recipes (apart from the original purpose of bootstrapping) as being somewhat similar to `.pyc` files in Python, or maybe more like WebAssembly -- a pre-compiled intermediate format. For that purpose, it's definitely thinking about whether the current approach is the best fit. I'd definitely be interested to hear about alternative ideas you have. I agree with Alex that it probably makes sense to stick with the current format until it causes problems and/or we have a better solution.

I'm not super concerned about people editing recipes directly. For now, I think it's ok for us to just say that doing so voids your warranty ;-)

I think adding the source to the recipe is an improvement to the situation, because then a recipe will be a superset of a text grammar. Then, the source always travels with the intermediate format, so the only way they can get out of sync is if people edit them directly.

pdubroy:
Ok, in my latest commit, it's intervals all the way down...

@alexwarth, if this looks good to you, I'll merge this and remove PExpr.prototype.toString() in a follow-up commit.

src/pexprs-outputRecipe.js
```diff
@@ -20,119 +20,103 @@ function escapeString(str) {
...
   sb.append(JSON.stringify(this.to));
-  sb.append(')');
+  sb.append(')' + getIntervalInfo(this, grammarInterval));
 };
 
-pexprs.Param.prototype.outputRecipe = function(sb, formals) {
-  sb.append('this.param(' + this.index + ')');
+pexprs.Param.prototype.outputRecipe = function(sb, formals, grammarInterval) {
+  sb.append('this.param(' + this.index + ')' + getIntervalInfo(this, grammarInterval));
+
```
mroeder: nit: blank line

mroeder:
LGMT (except for the blank line ;-))

src/GrammarDecl.js
```diff
@@ -21,6 +22,10 @@ function GrammarDecl(name) {
 
 // Helpers
 
+GrammarDecl.prototype.at = function(startIdx, endIdx) {
```
alexwarth: IMO this is method name makes call sites confusing, e.g.,

```.define("NameNoFormals", [], this.app("name").withInterval(decl.at(49, 53)))```

It might help to rename it to `sourceInterval`:

```.define("NameNoFormals", [], this.app("name").withInterval(decl.sourceInterval(49, 53)))```

File: src/GrammarDecl.js [hunk 3]
```diff
@@ -80,6 +85,11 @@ GrammarDecl.prototype.withDefaultStartRule = function(ruleName) {
   return this;
 };
 
+GrammarDecl.prototype.withSource = function(source) {
+  this.inputStream = InputStream.newFor(source);
```
alexwarth: It's strange for `GrammarDecl` to have an `InputStream` when what you really want is an `Interval`. Why not replace this line with:

```this.interval = new Interval(source, 0, source.length);```

File: src/GrammarDecl.js [hunk 4]
```diff
@@ -134,6 +144,10 @@ GrammarDecl.prototype.build = function() {
   if (grammarErrors.length > 0) {
     errors.throwErrors(grammarErrors);
   }
+  if (this.inputStream) {
+    grammar.definitionInterval = this.inputStream.interval(0, this.inputStream.source.length);
```
alexwarth: ... then this would become:

```grammar.definitionInterval = this.interval;```

File: src/GrammarDecl.js [hunk 5]
```diff
@@ -162,13 +176,14 @@ GrammarDecl.prototype.override = function(name, formals, body) {
   return this;
 };
 
-GrammarDecl.prototype.extend = function(name, formals, body) {
+GrammarDecl.prototype.extend = function(name, formals, fragment) {
```
alexwarth: Right, `body` wasn't quite right.

src/Interval.js
```diff
@@ -86,6 +87,19 @@ Interval.prototype = {
     }
   },
 
+  // Returns a new Interval that has the same extent as this one, but which is relative
+  // to `that`, an Interval that fully covers this one.
+  relativeTo: function(that, newInputStream) {
+    if (this.inputStream !== that.inputStream) {
+      throw errors.intervalSourcesDontMatch();
+    }
+    assert(this.startIdx >= that.startIdx && this.endIdx <= that.endIdx,
```
alexwarth: Why is this assertion not a proper error in `errors`, like `intervalSourcesDontMatch`, above? Is this something that can be caused by the user, or can it only happen as a result of a bug in the Ohm implementation?

src/pexprs.js
```diff
@@ -33,6 +35,14 @@ PExpr.prototype.withInterval = function(interval) {
   return this;
 };
 
+// Record the start and end index of the source for this expression, relative to an implicit
+// source string. Similar to `withInterval`, but can be used when it's not possible to pass
+// an Interval instance (e.g., inside a recipe).
+PExpr.prototype.fromSourceAt = function(startIdx, endIdx) {
```
alexwarth: I'm confused about this method. Does it really make sense to have an interval when you don't know the source string? And where is it being called?

test/test-recipes.js
```diff
@@ -56,13 +57,55 @@ test('recipes with supergrammars', function(t) {
...
+  t.equal(g.ruleBodies.ident.interval.contents, 'letter /* foo */ identPart*');
+
+  // Try re-parsing the grammar based on the source retained in the recipe.
+  var reconsitutedGrammar = ohm.grammar(g.definitionInterval.contents);
+  t.equal(reconsitutedGrammar.toRecipe(), ns.G.toRecipe());
+
+  var g2 = toGrammar(ns.G2.toRecipe());
+  t.equal(g2.ruleBodies.Start.interval.contents, '(ident | number)*', 'overridden rule');
+
+  // TODO: Figure out how to describe the interval for extended rules.
```
alexwarth: Shouldn't it just be the interval of the extended rule's declaration, i.e.,

  ruleName += ...

?

Were you thinking that it should also include the supergrammar's version of this rule? IMO we should be able to get to that, but it shouldn't be in the interval of this rule.

src/pexprs-outputRecipe.js
pdubroy: Whoops! Fixed.

src/GrammarDecl.js
pdubroy: Done.

src/GrammarDecl.js
pdubroy: Done.

src/Interval.js
pdubroy: It's the latter, basically. It should only if a recipe contains a rule whose interval (as specified by `.at(x, y)` is outside the bounds of the source that is included in that recipe. The only way that would happen is either (a) a bug, or (b) someone edited the recipe by hand (which they shouldn't do).

src/pexprs.js
pdubroy: Ah, good catch. This was left over from my initial implementation, which I wasn't happy with. Fixed.

test/test-recipes.js
pdubroy: You're right, it's an issue that we should deal with for `definitionInterval` (which the first version of this PR did) but not for `interval`. I think I just did `s/definitionInterval/interval` in this code, which makes this comment incorrect. Fixed.

src/GrammarDecl.js
pdubroy: I've changed it as you suggested. `at` was deliberately chosen to be short so as to not clutter the recipes -- `.withInterval(decl.sourceInterval())` is a lot of noise when it appears on every single expression. It's not a method that should really appear anywhere other than recipes, so I thought it would be better to make it short rather than descriptive.

But, it looks like we are going to revisit the recipes soon anyways, so I don't really care too much.

pdubroy:
Thanks for the comments -- I've addressed everything in the latest commit. @alexwarth, please take another look. If you think it looks good, I will squash the commits before merging.

alexwarth:
LGTM!

PR #70, opened 2016-02-27
toJson for Grammar and PExpr

tonyg:
The idea here is that one ought to be able to export a grammar using a plain data representation that can be read by *hypothetical* other implementations of Ohm.

For example,

```js
var ohm = require("ohm");
console.log(JSON.stringify(ohm.ohmGrammar.jsonHierarchy(), null, 2));
```

produces a JSON description of the core Ohm grammar.

I am finding this useful, for myself; is this a worthwhile thing to include in the package, for everybody?

(ETA: Comments and suggestions on style and functionality very welcome!)

### DIFF
--- a/src/Grammar.js
+++ b/src/Grammar.js
@@ -141,6 +141,54 @@ Grammar.prototype = {
     return false;
   },
 
+  toJson: function() {
+    var blob = {};
+    blob.name = this.name;
+    if (this.superGrammar) {
+      blob.superGrammar = this.superGrammar.name;
+    }
+    if (this.defaultStartRule) {
+      blob.defaultStartRule = this.defaultStartRule;
+    }
+    var definitions = {};
+    var extensions = {};
+    var overrides = {};
+    blob.definitions = definitions;
+    blob.extensions = extensions;
+    blob.overrides = overrides;
+    var self = this;
+    Object.keys(this.ruleBodies).forEach(function(ruleName) {
+      var body = self.ruleBodies[ruleName];
+      var formals = self.ruleFormals[ruleName];
+      var rule = {};
+      rule.name = ruleName;
+      rule.formals = formals;
+      rule.body = body.toJson(formals);
+      if (self.superGrammar && self.superGrammar.ruleBodies[ruleName]) {
+        (body instanceof pexprs.Extend ? extensions : overrides)[ruleName] = rule;
+      } else {
+        if (self.ruleDescriptions[ruleName]) {
+          rule.description = self.ruleDescriptions[ruleName];
+        }
+        definitions[ruleName] = rule;
+      }
+    });
+    return blob;
+  },
+
+  jsonHierarchy: function() {
+    var grammars = {};
+    var g = this;
+    while (g) {
+      grammars[g.name] = g.toJson();
+      g = g.superGrammar;
+    }
+    return {
+      startGrammar: this.name,
+      grammars: grammars
+    };
+  },
+
   toRecipe: function(optVarName) {
     if (this.isBuiltIn()) {
       throw new Error(
--- a/src/pexprs-toJson.js
+++ b/src/pexprs-toJson.js
@@ -1,0 +1,170 @@
+'use strict';
+
+// --------------------------------------------------------------------
+// Imports
+// --------------------------------------------------------------------
+
+var common = require('./common');
+var pexprs = require('./pexprs');
+
+// --------------------------------------------------------------------
+// Operations
+// --------------------------------------------------------------------
+
+pexprs.PExpr.prototype.toJson = common.abstract;
+
+pexprs.any.toJson = function(formals) {
+  return {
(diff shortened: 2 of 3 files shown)


src/Grammar.js
```diff
@@ -182,6 +182,54 @@ Grammar.prototype = {
     return false;
   },
 
+  toJson: function() {
+    var blob = {};
+    blob.name = this.name;
```
pdubroy: Nit: why not put this in the literal?

```
var blob = {name: this.name};
```

src/Grammar.js
```diff
@@ -182,6 +182,54 @@ Grammar.prototype = {
...
+    var extensions = {};
+    var overrides = {};
+    blob.definitions = definitions;
+    blob.extensions = extensions;
+    blob.overrides = overrides;
+    var self = this;
+    Object.keys(this.ruleBodies).forEach(function(ruleName) {
+      var body = self.ruleBodies[ruleName];
+      var formals = self.ruleFormals[ruleName];
+      var rule = {};
```
pdubroy: Same as above, I'd put these in the literal:

```
var rule = {
  name: ruleName,
  formals: formals,
  body: body.toJson(formals)
}
```

src/Grammar.js
```diff
@@ -182,6 +182,54 @@ Grammar.prototype = {
...
+  toJson: function() {
+    var blob = {};
+    blob.name = this.name;
+    if (this.superGrammar) {
+      blob.superGrammar = this.superGrammar.name;
+    }
+    if (this.defaultStartRule) {
+      blob.defaultStartRule = this.defaultStartRule;
+    }
+    var definitions = {};
```
pdubroy: I'd do this like so:

```
var definitions = blob.definitions = {};
```

...but I know that's not for everyone ;-)

src/pexprs-toJson.js
```diff
@@ -0,0 +1,170 @@
...
+};
+
+pexprs.Seq.prototype.toJson = function(formals) {
+  return {
+    type: 'seq',
+    factors: this.factors.map(function(f) { return f.toJson(formals); })
+  };
+};
+
+pexprs.Star.prototype.toJson = function(formals) {
```
pdubroy: You could share the same body on many of these, with something like this:

```
pexprs.Star.prototype.toJson =
pexprs.Plus.prototype.toJson = function(formals) {
  return {
    type: this.constructor.name.toLowerCase(),
    expr: this.expr.toJson(formals)
  };
}
```

pdubroy:
Hi Tony,

(Oops...somehow the comment I left here earlier wasn't submitted!)

Thanks for the PR! I was also wondering if we should consider something like this. One question - is part of the goal here to help with bootstrapping a new Ohm implementation? Because I'd consider the normal grammar syntax to be the host-language-independent representation.

Maybe we should replace the current "recipe" format with this JSON representation. See #69 for a recent discussion on recipes, if you didn't see it. One thing I'd like is for the JSON representation to include the original source, as in that PR (#69). The visualizer/editor is going to be a bigger focus for us in the future, and including the original source in the recipes (when it's available) makes for a better user experience.

alexwarth:
Right, I was also thinking that this JSON representation could be a good
replacement for recipes, and it that case, we would want some way to
include source code in them. (Seems relatively straightforward.)

Maybe the three of us should find a time to talk about this next week, over
Skype?

mroeder:
Hi Tony,
I thing it's nice to see the `toJson()` you did and I think what Alex and Pat suggest - replacing `toRecipe` with it - is possible. Two things come to my mind:

1. There is not yet a way to read the JSON back in and it won't be as fast as the recipes (but implementation-independent).
2. The additional information that Pat was looking for in #69 is not included in the JSON you create but for example in the CST (there is also an easy way to create JSON from the CST).

A similar JSON representation to what you are doing can currently be gotten from the `toAST` extra I put in:
```javascript
var ohm = require('ohm');
var toAST = require('ohm/extras').toAST;
var fs = require('fs');

var source = '' + fs.readFileSync('ohm/src/ohm-grammar.ohm'); // or any other Ohm grammar
var res = ohm.ohmGrammar.match(source);
var ast = toAST(res /*, mapping */);
console.log(JSON.stringify(ast, null, 2));
```

For a nicer representation, you can use the following `mapping`: https://gist.github.com/mroeder/b16fec1850690ec08a49
(`toAST` basically uses the [ESTree](https://github.com/estree) spec to create ASTs.)

pdubroy:
So, we talked about this, and agreed that it would be a good idea to move from the current recipe format to a JSON-based representation. Tony, is that something you'd be interested in implementing?

tonyg:
I really like the approach of pr #109. Accordingly, I'm closing this. Thanks all!

mroeder:
Thanks, Tony!
Let me know if you have any suggestions or comments on #109.

PR #72, opened 2016-03-08
Account for space skipping when looking for memoized trace entry.

pdubroy:
Since space skipping happens inside rule evaluation, the position used to look up a trace entry does not account for spaces. One solution is that when looking up the memoized trace entry, we use the memo table to skip over the space and find the entry we are looking for.

An alternative approach is to pull the space skipping logic up into `State.prototype.eval`. Here's a quick hack: https://github.com/cdglabs/ohm/commit/766d279703467981a9c0654045a49738911f82cd. The relevant tests seem to pass, so it doesn't seem obviously wrong.

@alexwarth, what do you think? I'm leaning towards the second option.


alexwarth:
Moving the space-skipping logic into `State.prototype.eval` seems like a good idea to me. I just don't understand the connection between that and what you did in 766d279. Is that just a temporary fix until we get around to changing `State.prototype.eval` as you explained?

pdubroy:
Yeah, sorry, that diff was a bit hard to read and includes some irrelevant stuff. The change to put space-skipping logic in `State.prototype.eval` was here: https://github.com/cdglabs/ohm/commit/766d279703467981a9c0654045a49738911f82cd#diff-c74d0b6564f4bf983d941d6a30952b22R272.

But yeah, that's not intended to be a proper fix. It was just a quick hack to get an idea of what changes would be required to get it to work. I thought it would be useful to include as context but maybe it was just confusing.

Anyways, thanks -- I'll prepare a proper patch to move the space skipping logic.

pdubroy:
@alexwarth, please take another look. I've left a few TODOs in here for things that I plan to address in a follow-up CL. One of them is a new `Match<startRule>` rule like we discussed -- this will allow us to move the `end` handling out of Apply, because it's a bit weird right now that `end` is a child of the top-level rule application.

Now that this is fresh in my brain, I'd also like to do a follow-up CL with some refactoring to make it easier to understand how tracing works with left recursion, but I'd rather land this first and to the refactoring in another commit if that's ok with you.

pdubroy:
Weird, I don't know what happened -- I tried to rebase the branch and somehow that automatically closed this PR. I think it should be fixed now though.

src/pexprs-eval.js
```diff
@@ -285,9 +280,11 @@ pexprs.Apply.prototype.eval = function(state) {
 
   var memoKey = app.toMemoKey();
   var memoRec = posInfo.memo[memoKey];
+  var isTopLevelApplication =
+      !caller && this !== state.applySpaces && this !== pexprs.end;
```
alexwarth: Nit: `pexprs.end` is not really an application, so it seems weird to have it show up in this expression. How about

    var isTopLevelApplication = !caller && this instanceof pexprs.Apply && this !== state.applySpaces;

instead?

alexwarth:
LGTM++

src/pexprs-eval.js
pdubroy: Right you are, not sure why I wrote that. Will fix in a follow-up commit tomorrow.

alexwarth: I just sent you a pull request -- see #80 :)

PR #75, opened 2016-03-11
Add CST node to traceNode to make all nodes in visualizer zoomable

lilyx:
To make all nodes zoomable in the visualizer, CST node have to be kept for applying semantic actions. So I added CST node to each of the succeed trace node, and some corresponding test. Doesn't seem to break anything. @alexwarth @pdubroy mind to take a look at it?


alexwarth:
Works for me. What do you think, @pdubroy?

BTW I think we should rename the `ans` argument of the `Trace` constructor to make it more descriptive. Before this change, maybe it should have been called `succeeded`? Now that it's either going to be a CST node or `false` to indicate a match failure, ... I'm not sure. Maybe `cstOrFalse` plus a comment that explains that it will be `false` if the match failed? Hopefully Pat will have a better suggestion :)

lilyx:
Made changes to the naming any styling issues base on the feedbacks, @alexwarth @pdubroy how about now?

PR #80, opened 2016-03-21
a little refactoring

alexwarth:
... that makes the variable name `isTopLevelApplication` more accurate in pexprs-eval.js. What do you think, @pdubroy?

### DIFF
--- a/src/pexprs-eval.js
+++ b/src/pexprs-eval.js
@@ -280,7 +280,7 @@ pexprs.Apply.prototype.eval = function(state) {
 
   var memoKey = app.toMemoKey();
   var memoRec = posInfo.memo[memoKey];
-  var isTopLevelApplication = !caller && this instanceof pexprs.Apply;
+  var isTopLevelApplication = !caller;
   return memoRec && posInfo.shouldUseMemoizedResult(memoRec) ?
       state.useMemoizedResult(memoRec) :
       app.reallyEval(state, isTopLevelApplication);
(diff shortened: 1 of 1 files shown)


src/pexprs-eval.js
```diff
@@ -280,8 +280,7 @@ pexprs.Apply.prototype.eval = function(state) {
 
   var memoKey = app.toMemoKey();
   var memoRec = posInfo.memo[memoKey];
-  var isTopLevelApplication =
-      !caller && this !== state.applySpaces && this !== pexprs.end;
+  var isTopLevelApplication = !caller && this instanceof pexprs.Apply;
```
pdubroy: `this` will always be an instance of pexprs.Apply, no?

pdubroy:
lgtm, though once you've moved the check for applySpaces, I think you can drop the `isTopLevelApplication` var and just pass `!caller`, which is what it did before (though keeping the var makes the code a bit more self-documenting).

pdubroy:
BTW, I still want to introduce the top-level `Match<startRule>` thing like we discussed, and this code will probably go away once I do that.

src/pexprs-eval.js
alexwarth: Oops! (*embarrassed*)

alexwarth:
Yes, I think that leaving that variable name there is nice b/c it's self-commenting, and (i) I agree that it still makes sense to introduce the `Match<startRule>` thing. Thanks!

PR #81, opened 2016-03-23
Refactoring the visualizer parts

mroeder:
... to create a public `ohmEditor` API and have only one global.

@pdubroy: How do you feel about that?
@lilyx: I know that this changes some things for the semantics editor as well, but my main goal was to refactor the code so I will have an easier way to programmatically load semantics. Please let me know any concerns and I am happy to work this refactoring into the changes on the `sv` branch!

### DIFF
--- a/test/test-visualizer.js
+++ b/test/test-visualizer.js
@@ -76,10 +76,11 @@ var HTML = '<div id="expandedInput"></div><div id="parseResults"></div>' +
 
 test('simple parse tree', function(t) {
   var doc = jsdom.jsdom(HTML);
-  var refreshParseTree = parseTree(ohm, doc, null, null);
+  var ohmEditor = {};
+  parseTree(ohm, ohmEditor, doc, null, null);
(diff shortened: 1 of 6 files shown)


pdubroy:
Hey Marko, this generally looks good. One thing I'd change is to only put stuff in `ohmEditor` that actually needs to be shared between modules. Some stuff (e.g. saveEditorState/restoreEditorState) should probably remain private to index.js. If we export everything, I'm not sure it provides much benefit over just using globals.

Also, when you merge, can you either squash the commits, or make sure that it's clear that they are all related to refactoring in the visualizer (e.g., by prepending `viz:` to them)? In the next Ohm release, I'm planning on producing a changelog, which means I'll have to scan through all the commits we've made in the past few months in order to provide a summary. That'll be easier if I can see at a glance what stuff is purely visualizer-related.

Anyways, LGTM if you trim the private stuff from `ohmEditor`.

PR #82, opened 2016-03-24
Visualizer: add buttons for adding semantic action & semantic editor

lilyx:
@pdubroy : First step for adding semantic actions to visualizer, only have the `add` and `action name` buttons, no semantic editor yet. Just want to let you review it before your vacation.


visualizer/index.html
```diff
@@ -165,6 +166,50 @@
...
+      resize: none;
+      outline: none;
+      padding-top: 1px;
+      padding-bottom: 1px;
+      margin-bottom: 4px;
+      height: 16px;
+      border: 1px solid #555;
+      border-radius: 10px;
+      display:block;
+      clear:both;
```
pdubroy: I don't think you need this. `clear` is usually only used with floated elements.

visualizer/index.js
```diff
@@ -140,6 +140,16 @@ function parseGrammar(source) {
     triggerRefresh(250);
   });
 
+  var actionContainers = document.querySelectorAll('.actionEntries');
+  Array.prototype.forEach.call(actionContainers, function(actionContainer) {
+    actionContainer.addEventListener('click', function() { triggerRefresh(); });
+    actionContainer.addEventListener('keypress', function(event) {
+      if (event.keyCode ===  13) {
```
pdubroy: Nit: looks like you've got an extra space in there before the `13`.

visualizer/index.html
```diff
@@ -46,6 +46,7 @@
     }
     #visualizerBody {
       flex-direction: column;
+      overflow: auto;
```
pdubroy: I don't think you'll need this anymore. As I was looking at this pull request (and trying to see why you needed to add the `overflow` here), I realized that the addition of the `#visualizerBody` in your earlier PR caused some problems with the scrolling. I've just pushed a fix, so I think you can remove this line from your patch now.

visualizer/index.html
```diff
@@ -165,6 +166,50 @@
       color: white;
     }
 
+    #actions {
+      display: flex;
+      flex: 1;
+      position: fixed;
```
pdubroy: I think this should be `position: absolute` now.

visualizer/index.html
```diff
@@ -165,6 +166,50 @@
...
+      right: 20px;
+      margin-top: 40px;
+    }
+    #actions .actionColumn {
+      margin-left: 15px;
+    }
+    #actions .actionEntries {
+      flex-direction: column;
```
pdubroy: You don't need this. `flex-direction` is a property for _flex containers_ (i.e., elements that have `display: flex`), not _flex items_. Here's a good guide to flexbox if you haven't seen it: https://css-tricks.com/snippets/css/a-guide-to-flexbox/

visualizer/index.html
```diff
@@ -165,6 +166,50 @@
...
+      align-self: flex-start;
```
pdubroy: I think you also might not need this. At least, when I disabled it, it didn't seem to change anything.

visualizer/index.js
```diff
@@ -140,6 +140,16 @@ function parseGrammar(source) {
     triggerRefresh(250);
   });
 
+  var actionContainers = document.querySelectorAll('.actionEntries');
```
pdubroy: Could you put the following comment above this?

```
// TODO: Move this code into parseTree.js.
```

This code is really specific to the parseTree, but I see that you need it here to be able to call `triggerRefresh`. I think @mroeder's refactoring is going to make stuff like this easier.

pdubroy:
LGTM with the comments I suggested.

visualizer/index.html
```diff
@@ -161,6 +171,105 @@
...
+    .semanticsEditor button.save {
+      width: 100%;
+    }
+    .semanticsEditor button.save:hover {
+      text-decoration: underline;
+    }
+
+    #actions {
+      display: flex;
+      position: fixed;
```
lilyx: Trying to keep the action buttons position fixed, so it won't change while we scrolling the bottom section

visualizer/index.html
```diff
@@ -161,6 +171,105 @@
...
+      background-color: rgba(0, 0, 0, 0);
+    }
+    textarea.action {
+      resize: none;
+      outline: none;
+      margin: 4px 4px 4px 0;
+      height: 16px;
+      border: 1px solid #999;
+      border-radius: 10px;
+      display:block;
```
lilyx: Missing a space, added in following commit

visualizer/index.html
```diff
@@ -203,7 +312,7 @@
     #bottomSection {
       display: flex;
       flex: 1;
-      flex-direction: column;
```
lilyx: Trying to list zoomOutButton, parse result, and action buttons side by side, so there won't be too much vertical spaces taken by bottoms.

visualizer/index.html
```diff
@@ -203,7 +312,7 @@
...
+      overflow-x: auto;
```
lilyx: Scroll horizontally for long expanded example.

src/Semantics.js
```diff
@@ -380,6 +388,17 @@ Semantics.prototype.extendOperationOrAttribute = function(type, name, actionDict
   this[typePlural][name].checkActionDict(this.grammar);
 };
 
+Semantics.prototype.getOperationOrAttribute = function(operationOrAttributeName) {
+  if (operationOrAttributeName in this.attributes) {
+    return this.attributes[operationOrAttributeName];
+  } else if (operationOrAttributeName in this.operations) {
+    return this.operations[operationOrAttributeName];
+  } else {
+    throw new Error('Cannot find operation or attribute has name ' +
```
pdubroy: I would just return undefined here, and let the calling code handle the error.

visualizer/index.html
```diff
@@ -161,6 +171,105 @@
       color: white;
     }
 
+    .semanticsEditor.hidden .header,
+    .semanticsEditor.hidden .body,
+    .semanticsEditor.hidden .save {
+      display: none;
```
pdubroy: Can you just give 'display: none' to the entire .semanticsEditor?

visualizer/index.html
```diff
@@ -161,6 +171,105 @@
...
+    #actions .actionEntries {
+      display: flex;
+      flex-direction: column;
+      background-color: white;
+      color: black;
+      font-family: Menlo, Monaco, sans-serif;
+      font-style: normal;
+      border-radius: 3px;
+      padding: 2px 0px 2px 0px;
+      background-color: rgba(0, 0, 0, 0);
```
pdubroy: Duplicate entry of `background-color`.

visualizer/index.html
pdubroy: There's something not quite right here. With a big example, I now see two horizontal scroll bars in the bottom pane.

visualizer/index.html
lilyx: For the editor that has classname 'hidden' I want to show the result only. I could see that 'hidden' is a confusing classname here, maybe change to 'resultOnly'?

visualizer/index.html
lilyx: Couldn't reproduce the problem. Could you specify which example you are using?

PR #84, opened 2016-03-29
Visualizer: add semantic action buttons

lilyx:
Revert back to adding semantic action button stage. Not exactly the same as before, and want to make sure it's okay to merge this part.


visualizer/actionButton.js
```diff
@@ -0,0 +1,71 @@
...
+(function() {
+
+  function selectNode(actionNode) {
+    actionNode.classList.add('selected');
+    if (!actionNode.readOnly) {
+      actionNode.select();
+    }
+  }
+
+  function addAction(actionType) {
```
pdubroy: This code is very nice and readable! Good job with formatting and naming.

In the future, you might want to add at least a few short comments describing the most important functions, and/or a comment describing the purpose of the file. Even when the code is nice and simple like this, the comments give a good starting point for reading the code, and makes it easier to have an overview of the code at a glance.

visualizer/index.html
```diff
@@ -161,6 +160,46 @@
...
+    }
+    #actions .actionEntries {
+      display: flex;
+      flex-direction: column;
+      color: black;
+      font-family: Menlo, Monaco, sans-serif;
+      font-style: normal;
+      border-radius: 3px;
+      padding: 2px 0px 2px 0px;
+      background-color: rgba(0, 0, 0, 0);
```
pdubroy: Nit: why not just make this `background-color: black`?

visualizer/index.html
```diff
@@ -203,7 +242,7 @@
     #bottomSection {
       display: flex;
       flex: 1;
-      flex-direction: column;
+      overflow-x: auto;
```
pdubroy: This still produces double horizontal scroll bars for me. I'm using the ES5 grammar, with the following input: `var x = "asdlfkjasdlfkjasdf";`. The parse tree is big enough to trigger scrolling -- on master, I just see a single horizontal scroll bar, but on this branch, I see two.

Currently, the way the scrolling works is this: `#bottomSection` and `#visualizerBody` do not scroll -- only `#parseResults` does. I did that because I wanted the expanded input to always be visible, and not scroll out of view.    But the problem with that is that we want to be able to scroll the expanded input horizontally, while still keeping it visible when the parse tree is scrolled vertically. So I see what you're trying to do here.

What could work is if we make `#bottomSection` scroll _only_ horizontally (like you've done here), and I change `#parseResults` to scroll only vertically. That will eliminate the double scroll bar problem, and should give us the behaviour that we want.

visualizer/parseTree.js
```diff
@@ -237,35 +237,14 @@
...
-    refreshParseTree(ui, grammar, rootTrace, showFailures, {zoomTrace: traceNode});
+    return currentRootTrace !== traceNode &&
+           traceNode.succeeded &&
+           !isPrimitive(traceNode.expr) &&
+           traceNode.expr.ruleName !== 'spaces';
   }
 
   function createTraceElement(ui, grammar, rootTrace, traceNode, parent, input,
-    showFailures, optZoomState) {
+    optActionName, optZoomState) {
```
pdubroy: Nit: when continuing a line like this, it should be indented an extra two spaces (we can fix in a follow-up commit).

pdubroy:
LGTM. I'm going to go ahead and merge so I can fix up the scroll bar thing and a few other small issues.

visualizer/index.html
pdubroy: Ok, I just pushed a change that implements what I described above. It's not quite right, because the vertical scrollbar is only visible when `#bottomSection` is scrolled all the way to the right, but I think that's fine for now.

It's a really tricky layout, and we'll have to think about whether there's a better way to solve it.

PR #86, opened 2016-03-30
Visualizer: Add semantics editor

lilyx:
Adding semantics editor to the visualizer. Hope this one makes review easier :)


pdubroy:
Thanks Meixian, that's perfect! I think Alex is going to take a look at this with you today, and I'll follow up tomorrow with my comments.

visualizer/parseTree.js
```diff
@@ -156,6 +159,31 @@
     }
   }
 
+  // Hides or shows the semantic editor of `el`, which is a div.pexpr.
```
alexwarth: s/semantic/semantics

visualizer/parseTree.js
```diff
@@ -156,6 +159,31 @@
...
+  function toggleSemanticEditor(el) {
```
alexwarth: Please add a comment that explains what this does, in a way that someone using it would understand.

pdubroy:
I think this is out of date, right?

PR #87, opened 2016-03-30
Add `toArgumentNameList` method

lilyx:
@alexwarth refactoring of the `toArgumentString` method and rename it to `toArgumentNameList`.

### DIFF
--- a/src/pexprs-toArgumentNameList.js
+++ b/src/pexprs-toArgumentNameList.js
@@ -1,0 +1,149 @@
+'use strict';
+
+// --------------------------------------------------------------------
+// Imports
+// --------------------------------------------------------------------
+
+var common = require('./common');
+var pexprs = require('./pexprs');
+
+var copyWithoutDuplicates = common.copyWithoutDuplicates;
+
+// --------------------------------------------------------------------
+// Operations
+// --------------------------------------------------------------------
+
+/*
+  Returns a list of strings that will be used as the default argument names for its receiver
+  (a pexpr) in a semantic action. This is used exclusively by the Semantics Editor.
+
+  `firstArgIndex` is the 1-based index of the first argument name that will be generated for this
+  pexpr. It enables us to name arguments positionally, e.g., if the second argument is a
+  non-alphanumeric terminal like "+", it will be named '$2'.
+
+  Here is a more elaborate example that illustrates how this method works:
+  `(a "+" b).toArgumentNameList(1)` evaluates to `['a', '$2', 'b']` with the following recursive
+  calls:
+
+    (a).toArgumentNameList(1) -> ['a'],
+    ("+").toArgumentNameList(2) -> ['$2'],
+    (b).toArgumentNameList(3) -> ['b']
+
+  Notes:
+  * This method must only be called on well-formed expressions, e.g., the receiver must
+    not have any Alt sub-expressions with inconsistent arities.
+  * e.getArity() === e.toArgumentNameList(1).length
+*/
+pexprs.PExpr.prototype.toArgumentNameList = common.abstract;  // function(firstArgIndex) { ... }
+
+pexprs.any.toArgumentNameList = function(firstArgIndex) {
+  return ['any'];
+};
+
+pexprs.end.toArgumentNameList = function(firstArgIndex) {
+  return ['end'];
+};
+
+pexprs.Prim.prototype.toArgumentNameList = function(firstArgIndex) {
+  if (typeof this.obj === 'string' && /^[_a-zA-Z0-9]+$/.test(this.obj)) {
+    // If this terminal is a valid suffix for a JS identifier, just prepend it with '_'
+    return ['_' + this.obj];
+  } else {
+    // Otherwise, name it positionally.
+    return ['$' + firstArgIndex];
+  }
+};
+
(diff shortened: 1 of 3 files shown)


src/pexprs-toArgumentNameList.js
```diff
@@ -0,0 +1,98 @@
...
+
+/*
+  The `toArgumentNameList` method returns a list of strings that will
+  be used as the default argument names for its receiver (a pexpr) in a
+  semantic action. This is used exclusively by the Semantics Editor.
+*/
+
+pexprs.PExpr.prototype.toArgumentNameList = common.abstract;
+
+pexprs.Prim.prototype.toArgumentNameList = function(optIndex) {
```
alexwarth: Is the argument here really optional? It doesn't seem to be, since it's used in the `else` branch below.

src/pexprs-toArgumentNameList.js
```diff
@@ -0,0 +1,98 @@
...
+  } else {
+    return ['$' + optIndex];
+  }
+};
+
+pexprs.Range.prototype.toArgumentNameList = function(optIndex) {
+  return [this.from + '_to_' + this.to];
+};
+
+pexprs.Alt.prototype.toArgumentNameList = function(optIndex) {
```
alexwarth: It would be nice to have some comments explaining what each "paragraph" inside this method is doing.

src/pexprs-toArgumentNameList.js
```diff
@@ -0,0 +1,98 @@
...
+  var termArgumentNames = this.terms.map(function(term, idx) {
+    return term.toArgumentNameList(idx + 1);
```
alexwarth: I'm pretty confused about how these indices are used -- it would be good to write a comment about them, too.

src/pexprs-toArgumentNameList.js
```diff
@@ -0,0 +1,98 @@
...
+  });
+
+  return ans.map(function(argumentList) {
+    return argumentList.join('_or_');
+  });
+};
+
+pexprs.Seq.prototype.toArgumentNameList = function(optIndex) {
+  var ans = [];
+  var count = {};
```
alexwarth: Nit: you should change `{}` to `Object.create(null)`, or else names like `toString` won't do the right thing.

src/pexprs-toArgumentNameList.js
```diff
@@ -0,0 +1,98 @@
...
+
+  return ans;
+};
+
+pexprs.Star.prototype.toArgumentNameList =
+pexprs.Plus.prototype.toArgumentNameList = function(optIndex) {
+  var exprArgumentStrings = this.expr.toArgumentNameList(optIndex);
+
+  return exprArgumentStrings.map(function(exprArgumentString) {
+    return exprArgumentString + 's';
```
alexwarth: To be cute, we may want to append `'es'` instead of `'s'` if `exprArgumentString` ends with `'s'`. Then:

* `(Expr*).toArgumentNameList()` will evaluate to `['Exprs']`, and
* `(Exprs*).toArgumentNameList()` will evaluate to `['Exprses']`.

src/pexprs-toArgumentNameList.js
```diff
@@ -0,0 +1,98 @@
...
+  });
+};
+
+pexprs.Opt.prototype.toArgumentNameList = function(optIndex) {
+  return ['opt' + this.expr.toArgumentNameList(optIndex)];
```
alexwarth: Nit: I think we should capitalize the first letter of `this.expr.toArgumentNameList()`, if it is indeed a letter. E.g., `(keyword?).toArgumentNameList()` -> `['optKeyword']`.

src/pexprs-toArgumentNameList.js
```diff
@@ -0,0 +1,98 @@
...
+      return;
+    }
+  });
+
+  return ans;
+};
+
+pexprs.Star.prototype.toArgumentNameList =
+pexprs.Plus.prototype.toArgumentNameList = function(optIndex) {
+  var exprArgumentStrings = this.expr.toArgumentNameList(optIndex);
```
alexwarth: IMO the body of this method would be more readable if you used a different name for the argument of the anonymous function. What do you think about this:

```
....prototype.toArgumentList= function(optIndex) {
  return this.expr.toArgumentNameList(optIndex).map(function(argumentName) {
    return argumentName + 's';
  });
};
```

?

src/pexprs-toArgumentNameList.js
```diff
@@ -0,0 +1,98 @@
...
+          if (count[factorArgumentName] === 1) {
+            ans[ans.indexOf(factorArgumentName)] = factorArgumentName + '_1';
+          }
+          factorArgumentName = factorArgumentName + '_' + (++count[factorArgumentName]);
+        }
+
+        ans.push(factorArgumentName);
+      });
+    } catch (error) {
+      return;
```
alexwarth: Why do we need a try/catch in this method? Under what conditions do you expect there to be an error? Do you really want to return `undefined` when that happens? (I don't think so, because it will cause recursive calls to break.)

alexwarth:
I game @lilyx a bunch of feedback offline, and I'm really happy with the latest version. So merging.

PR #88, opened 2016-03-31
Recipes for Semantics

mroeder:
Here is a WIP for recipes for semantics. It's making use of the recipes for grammars and bundles both in a similar style (though only the semantics will be returned and the grammar can be retrieved from it).

Sketch of the resulting recipe:
```javascript
(function() {
  var grammarFn =[whatever grammar.toRecipe() returns];
  var grammar = grammarFn.call(this);
  var semantics = grammar.semantics()
    .addOperation("foo", {
       // all the actions
    })
    .addOperation("bar<x>", {
      // and another operation
    })
    .addAttribute("baz", {
      // same for attributes
    });
  return semantics;
})
```

There are still some `FIXME`s in there that obviously will have to go before merging (**DO NOT COMMIT YET**).

@pdubroy @alexwarth, I am looking for feedback and whether this is what you had in mind as well as hints for what would be a good case for *extended semantics* (not looked at yet).


src/Semantics.js
```diff
@@ -196,6 +196,41 @@ Semantics.prototype.checkActionDicts = function() {
   }
 };
 
+Semantics.prototype.toRecipe = function() {
+  var sb = new common.StringBuffer();
```
pdubroy: BTW, I'm relatively sure that this StringBuffer technique doesn't give any performance benefit in modern VMs. I've been meaning to go through and remove it everywhere. So if you want to avoid using it here, I'd be in favour :-)

src/Semantics.js
```diff
@@ -196,6 +196,41 @@ Semantics.prototype.checkActionDicts = function() {
...
+  sb.append('(function() {\n  var grammarFn = ');
+  sb.append(this.grammar.toRecipe());
+  sb.append('  var grammar = grammarFn.call(this);\n');
+  sb.append('  var semantics = grammar.semantics()');
+
+  ['Operation', 'Attribute'].forEach(function(type) {
+    var semantics = this[type.toLowerCase() + 's'];
```
pdubroy: This name seems not quite right, b/c this variable doesn't hold a semantics instance. It's a list of all operations (resp. attributes), right?

src/Semantics.js
```diff
@@ -196,6 +196,41 @@ Semantics.prototype.checkActionDicts = function() {
...
+    Object.keys(semantics).forEach(function(semName) {
```
pdubroy: ...and I would just call this `name`, not `semName`.

src/Semantics.js
```diff
@@ -196,6 +196,41 @@ Semantics.prototype.checkActionDicts = function() {
...
+      var fullName = semName;
```
pdubroy: Maybe name this `signature`?

src/Semantics.js
```diff
@@ -196,6 +196,41 @@ Semantics.prototype.checkActionDicts = function() {
...
+      if (semantics[semName].formals.length > 0) {
+        fullName += '(' + semantics[semName].formals.join(', ') + ')';
+      }
+      sb.append('\n    .add' + type + '(' + JSON.stringify(fullName) + ', {');
+
+      var actions = semantics[semName].actionDict;
+      var followUp = false;
+      Object.keys(actions).forEach(function(actionName) {
+        if (actionName === '_default') {
+          return; // FIXME: check whether _default has been overwritten
```
pdubroy: Maybe you could just check if it's `===` to the _default_ `_default`?

src/Semantics.js
```diff
@@ -196,6 +196,41 @@ Semantics.prototype.checkActionDicts = function() {
...
+        }
+        sb.append(followUp ? ',' : (followUp = true, ''));
```
pdubroy: I find this to be really tricky. Assuming I understand it correctly, it might be more straightforward to do it like this: `var arr = Object.keys(actions).map(function(actionName) { ... });`, followed by `arr.join(',')`?

test/test-recipes.js
```diff
@@ -12,7 +12,7 @@ var testUtil = require('./testUtil');
 // Helpers
 // --------------------------------------------------------------------
 
-function toGrammar(recipeString) {
+function toGrammarOrRecipe(recipeString) {
```
pdubroy: Maybe just call this `makeRecipe`? Or else it should be `toGrammarOrSemantics` I think.

pdubroy:
Looks pretty good to me.

src/Semantics.js
alexwarth: +1

src/Semantics.js
alexwarth: Right. Maybe rename it `operationsOrAttributes`?

src/Semantics.js
alexwarth: +1

src/Semantics.js
alexwarth: Ooo, this gives me an idea about the operations-and-attributes grammar. Thanks!

src/Semantics.js
mroeder: I was wondering whether there is a nice word for that - despite o-pe-ra-tions-and-a-ttri-butes. ;-)
I checked the documentation but it doesn't exist. :-( Maybe there will also be more in the future, `operationsAndAttributesAndXAndY`... Can we can up with a nicer, general name? (I also think that Meixian had a similar problem when writing code for the semantics editor...)

src/Semantics.js
mroeder: I thought about it. And wanted to untangle it. But at the same time I also thought that this would make it "unique" (Singleton-like) and any changes to the *default* `_default` would be made to all grammars. (Sure, functions are pseudo-immutable and usually you replace `_default` it in your semantic - if necessary at all -, but who knows...)

test/test-recipes.js
mroeder: Yikes! So obvious (it should have been `toGrammarOrSemantics`). But I will call it `makeRecipe`. Thanks :-)

src/Semantics.js
pdubroy: I agree, it's cumbersome and confusing to use two different names for things that are so similar. I think right now there are only two differences:

1. Attributes are memoized, and
2. Attributes can't have arguments

...and I could imagine that at some point we would support memoizing with arguments (though it's not clear exactly how it would work).

I'd be in favour of just calling everything "semantic operations", though using "attributes" has the advantage of being connected to attribute grammars.

src/Semantics.js
mroeder: Wow, it seems to be even worse:
```javascript
console.time('string');
var str = '';
for (var i = 0; i < 1000000; i++) {
  str += 'The line with ' + i + '\n';
}
var all = str;
console.timeEnd('string');
```
`node performance1.js` done 10x: 173ms+/-2ms

```javascript
console.time('buffer');
var str = [];
for (var i = 0; i < 1000000; i++) {
  str.push('The line with ' + i);
}
var all = str.join('\n');
console.timeEnd('buffer');
```
`node performance2.js` done 10x: 316ms+/-10ms

Add the overhead of `new` and method calls...

Is that what those little examples show what you mean here?

src/Semantics.js
mroeder: ... and there is a difference in how I use it: with or without `()`

src/Semantics.js
alexwarth: Yeah, IIRC that's the kind of difference in performance that Pat saw when he measured both approaches. The funny thing is that when I started implementing Ohm, I went with the `StringBuffer` approach b/c I did a comparison in jsperf and found out that it was significantly faster than using `+=`. But that was then, and this is now.

This is one of those cases where it would be really nice to have PL support for separating meanings from optimizations...

pdubroy: Beware of microbenchmarks...you sure the JIT didn't eliminate all the code between the `console.time()` and `console.timeEnd()`? ;-)

mroeder: Now... yes! :-p (added a `console.error(all)` and `2> /dev/null`)
But I was thinking that it might make a difference in a bigger context - yes.
(https://jsperf.com/string-concat-vs-array-join-10000/)

test/test-recipes.js
```diff
@@ -110,7 +110,100 @@ test('recipes with source', function(t) {
...
+    .addAttribute('value', {
+      number: function(digits) {
+        return parseFloat(this.interval.contents);
+      }
+    });
+
+  var s2 = makeRecipe(s1.toRecipe());
+  t.equal(s2.name, 'ASemantics', 'semantics created');
+
+  var g2 = s2._getSemantics().grammar; // FIXME: should have an official API
```
mroeder: @alexwarth @pdubroy Should I leave this like that for now or would it be good to add a *public* `getGrammar` to the semantics wrapper?

test/test-recipes.js
```diff
@@ -110,7 +110,100 @@ test('recipes with source', function(t) {
...
+  t.equal(g2.name, 'G', 'semantics from grammar');
+
+  var ops = s2._getSemantics().operations; // FIXME: should have an official API
```
mroeder: @alexwarth @pdubroy What about this one? Should there be a way to get all the (names of) operations and attributes? (This could be paired with Meixian's `getOperationOrAttribute`.)

test/test-recipes.js
alexwarth: I suppose adding a `getGrammar` method to the `Semantics` wrappers would be OK. But what I think we really want is an object that represents an "ohm bundle" that would have a grammar and a semantics, as well as examples and tests to go with those. (These examples and tests should probably be included in their associated grammar and semantics, instead of being part of the "bundle".) The three of us should talk about this soon.

test/test-recipes.js
alexwarth: I think this is a good idea. How about `getOperationNames` and `getAttributeNames`?

mroeder:
The last change (aed4518)  is not as nice as I would want it to be. Please, leave feedback if you have a better idea!

src/Semantics.js
```diff
@@ -246,6 +310,42 @@ function parsePrototype(nameAndFormalArgs, allowFormals) {
   return prototypeGrammarSemantics(r).parse();
 }
 
+function newDefaultAction(typePlural, type, name, doIt) {
```
pdubroy: I think it would be slightly clearer to just pass in the Operation/Attribute here, instead of `typePlural`.

src/Semantics.js
```diff
@@ -260,39 +360,7 @@ Semantics.prototype.addOperationOrAttribute = function(type, nameAndFormalArgs,
...
-        // applying this operation / attribute to the child node.
-        return doIt.apply(children[0], args);
-      } else {
-        // Otherwise, we throw an exception to let the programmer know that we don't know what
-        // to do with this node.
-        throw new Error(
-            'Missing semantic action for ' + this.ctorName + ' in ' + name + ' ' + type);
-      }
-    }
+    _default: newDefaultAction(typePlural, type, name, doIt)
```
pdubroy: Instead of the string comparison, what if you just save the result of `newDefaultAction`, and either pass that in to the Operation/Attribute constructor, or just set it as a property later?

E.g.:
```
var builtInDefault = newDefaultAction(...);
var realActionDict = {_default: builtInDefault};
...
this[typePlural][name] = type === 'operation' ?
       new Operation(name, inheritedFormals, newActionDict, builtInDefault) :
       new Attribute(name, newActionDict, builtInDefault);
```
Then you can use simple equality to ensure that you don't include that in the recipe.

src/Semantics.js
```diff
@@ -196,6 +196,70 @@ Semantics.prototype.checkActionDicts = function() {
...
+      if (hasSuperSemantics(this) && this.super[type.toLowerCase() + 's'][name]) {
+        method = 'extend' + type;
+      } else {
+        method = 'add' + type;
+      }
+      str += '\n    .' + method + '(' + JSON.stringify(signature) + ', {';
+
+      var actions = semanticOperations[name].actionDict;
+      var defaultStr = newDefaultAction(/* can omit params */).toString();
+      var srcArray = Object.keys(actions).filter(function(actionName) {
```
pdubroy: IMO, this would be slightly clearer as a `forEach` with an `if` statement inside, but that's just a personal preference.

src/Semantics.js
```diff
@@ -451,6 +519,18 @@ Semantics.createSemantics = function(grammar, optSuperSemantics) {
     s.extendOperationOrAttribute.call(s, 'attribute', name, actionDict);
     return proxy;
   };
+  proxy.getOperationNames = function() {
```
pdubroy: Hmmm, I wonder if we should just be exposing the operations and attributes themselves...

I'm fine with this for now, just something for us to think about in the future.

pdubroy:
LGTM apart from what I said about the toString() thing...I think it's also merge this and fix in a follow-up commit.

And can you squash if/when you merge, pretty please :-)

src/Semantics.js
mroeder: I was actually thinking that this is the better thing to do (but was pondering too after Alex suggested this). You might want to know which operations/attributes a semantics supports but you will not access them but rather call them through `s(matchResult).[operationOrAttributeName](...)`.

PR #89, opened 2016-04-01
Add API to prepare for adding semantics editor

lilyx:


src/Semantics.js
```diff
@@ -451,6 +459,14 @@ Semantics.createSemantics = function(grammar, optSuperSemantics) {
     s.extendOperationOrAttribute.call(s, 'attribute', name, actionDict);
     return proxy;
   };
+  proxy.getActionDict = function(operationOrAttributeName) {
```
alexwarth: This is not part of the "public" API, so let's prepend the name with a `_`.

src/Semantics.js
```diff
@@ -451,6 +459,14 @@ Semantics.createSemantics = function(grammar, optSuperSemantics) {
...
+    var action = s.operations[operationOrAttributeName] || s.attributes[operationOrAttributeName];
+    if (!action) {
+      throw new Error('There is no action has name ' + operationOrAttributeName +
```
alexwarth: How about `operationOrAttributeName + ' is not a valid operation or attribute name in this semantics'`?

alexwarth:
I just added a couple of small comments. This looks great, it's just missing a unit test.

lilyx:
Updated with unit test.

PR #90, opened 2016-04-06
Visualizer: Fix of updating zoomState when input/grammar changes

lilyx:
`zoomState` needs to be cleared when input/grammar changed


pdubroy:
LGTM, but after the demo we should think about whether there is a better (or more general) way to do this.

PR #96, opened 2016-05-03
Build on windows successfully

justinmchase:
When npm runs on windows (even from git bash) it executes scripts as CMD
scripts. Using `$var` in these scripts causes issues on Windows.

Bash can be invoked however, so moving these scripts into shell files in
the bin directory allows the scripts to be completely bash while still
working on windows.

### DIFF
--- a/bin/build-debug.sh
+++ b/bin/build-debug.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+if [ $npm_package_browserify_options ]; then
+  options=$npm_package_browserify_options
+else
+  options=$NPM_PACKAGE_BROWSERIFY_OPTIONS
+fi
+
+browserify $options
--- a/bin/build-debug.sh
+++ b/bin/build-debug.sh
@@ -1,8 +1,4 @@
 #!/bin/bash
-if [ $npm_package_browserify_options ]; then
-  options=$npm_package_browserify_options
-else
-  options=$NPM_PACKAGE_BROWSERIFY_OPTIONS
-fi
+options=${NPM_PACKAGE_BROWSERIFY_OPTIONS:-$npm_package_browserify_options}
 
 browserify $options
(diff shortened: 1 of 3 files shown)


pdubroy:
Hi Justin, thanks for this! Happy to make the changes required to get this stuff working on Windows.

bin/build-debug.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+if [ $npm_package_browserify_options ]; then
+  options=$npm_package_browserify_options
+else
+  options=$NPM_PACKAGE_BROWSERIFY_OPTIONS
```
pdubroy: Can you explain why you need to check both here? I know a lot of stuff on Windows is case-insensitive...is NPM setting these vars in uppercase? Do you think we can  be assured that only one of them is set?

Might be cleaner to do it like this:

```
# Account for both Unix (lowercase) and Windows (uppercase).
OPTIONS=${npm_package_browserify_options:−$NPM_PACKAGE_BROWSERIFY_OPTIONS}
```

package.json
```diff
@@ -38,15 +38,16 @@
...
+    "watch": "bash bin/watch.sh"
   },
   "license": "MIT",
   "author": "Alex Warth <alexwarth@gmail.com> (http://tinlizzie.org/~awarth)",
   "contributors": [
     "Patrick Dubroy <pdubroy@gmail.com>",
     "Tony Garnock-Jones <tonygarnockjones@gmail.com>",
     "Yoshiki Ohshima <Yoshiki.Ohshima@acm.org>",
-    "Marko Röder <m.roeder@photon-software.de>"
+    "Marko Röder <m.roeder@photon-software.de>",
```
pdubroy: Sorry, would you mind leaving this out of this PR? Totally happy to give credit to everyone who has contributed, but right now this list is kind of arbitrary...and rather than add only you, I think we should just add everyone who has contributed, or drop this from the package.json. Either way, I'd prefer to handle in a separate PR.

bin/build-debug.sh
justinmchase: Sure, I like that. For whatever reason they get capitalized when going through CMD and bash is case sensitive. Pretty evil.

package.json
justinmchase: Sure.

justinmchase:
@pdubroy Fixed!

pdubroy:
Thanks! Merged.

Issue #94, opened 2016-05-03
The ecmascript package example can't work


justinmchase:
Due to this line:
https://github.com/cdglabs/ohm/blob/master/examples/ecmascript/es5.js#L12

When its installed via package it is no longer a subdirectory of ohm. I think if you just `require('ohm-js')` it should resolve up the directory chain.

But given its a peer dependency it seems like it should probably change it to allow ohm to be passed in via peer also, something like:

```
var ohm = require('ohm-js');
var es = require('ohm-example-ecmascript');
var es5 = es.es5({ ohm: ohm });
...
```

pdubroy:
Ack, thanks for pointing this out. I published that package rather hastily, and clearly didn't test it properly (or at all?).

Thanks for submitting a PR for this (#97).

PR #97, opened 2016-05-03
Ecmascript grammar accepts ohm

justinmchase:
- [ ] merge #96 

I was unable to use the ecmascript module right out of npm. This change allows it to be used and it exposes es5 and es6 with the same api.

This PR is dependent on PR #96 being merged first.

Requiring the module now loads index.js and the api is a little different, therefore I recommend bumping the major version of this library.

### DIFF
--- a/bin/build-debug.sh
+++ b/bin/build-debug.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+if [ $npm_package_browserify_options ]; then
+    options=$npm_package_browserify_options
+else
+    options=$NPM_PACKAGE_BROWSERIFY_OPTIONS
+fi
+
+browserify $options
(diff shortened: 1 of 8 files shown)


pdubroy:
Thanks for this PR. I definitely want to fix the ecmascript module, but I want to put a bit more thought into how we want to handle this in general.

In the meantime, any package that is already depending on Ohm can already use the ES5 grammar like this: `var es5 = require('ohm-js/examples/ecmascript/es5')`

justinmchase:
Ok thanks for the workaround. Do you want me to close this pr?

pdubroy:
Yeah, might be best to close it for now -- it's covered by #94.

justinmchase:
I'll leave the branch alive for a little while in case you want to look at it. Thanks.

PR #100, opened 2016-05-04
Fixing parsing example

jogleasonjr:
The HTTPS resource in JSFiddle was broken. Changed it to HTTP (not ideal long term). And there was no "result" element to pin the results to.


pdubroy:
Hi John,

Thanks for pointing this out. I fixed it in my fiddle directly.

PR #101, opened 2016-05-04
fix "result" pre element that was pasted over

jogleasonjr:


pdubroy:
Fixed in the original JSFiddle.

PR #102, opened 2016-05-06
Remove support for matching on structured data

pdubroy:
Remove support for all expression types that match on non-string data: Arr, Obj, Str, numbers, and keywords (true, false, null).


NikolaySuslov:
Hello,

From the beginning of Ometa the support for matching on structured data was
one of the gorgeous and unique language features. Is there any other
modular way of preserving this feature in Ohm, while not just removing it
completely from the core?

Regards,
Nikolay

PR #104, opened 2016-05-19
Visualizer: add action buttons for semantics

lilyx:


test/test-visualizer.js
```diff
@@ -101,7 +101,8 @@ test('simple parse tree', function(t) {
   ohmEditor.registerEvents({
     'parse:input': ['matchResult', 'trace'],
     'change:inputEditor': ['codeMirror'],
-    'change:grammarEditor': ['codeMirror']
+    'change:grammarEditor': ['codeMirror'],
+    'change:action': ['trace']
```
pdubroy: Is this something that effects other parts of the editor? If so, then it's fine to put it here. But if it's specific to the semantics editor, then it probably belongs somewhere else. E.g., maybe it makes sense on `ohmEditor.parseTree`...or maybe we should create create a new CheckedEmitter named `ohmEditor.semantics`.

And it would be good to have a bit more descriptive of a name...with no other context, "action" is pretty ambiguous. What about "change:semanticOperation" (isn't it the _operation_ that's changing)?

visualizer/actionButton.js
```diff
@@ -0,0 +1,265 @@
...
+  ohmEditor.addListener('parse:input', rebootRootTraceForSemantics);
+
+  var UnicodeChars = {
+    BLACK_UP_POINTING_TRIANGLE: '\u25B2',
+    BLACK_DOWN_POINTING_TRIANGLE: '\u25BC',
+    PLUS_SIGN: '\u002B',
+    LEFTWARDS_ARROW: '\u2190'
+  };
+
+  function createActionNameContainer() {
```
pdubroy: I think everything you're calling "action" in this file is really an "operation", right? I think the terminology we've been using is that "semantic action" means a function for a specific node type, and "operation" is `toAST()`, `value`, etc.

visualizer/index.html
```diff
@@ -398,6 +398,123 @@
       display: none;
     }
 
+    #addOperation,
```
pdubroy: I think it would be good to put all this stuff into a separate .css file named "semanticsEditor.css". I've been meaning to pull the other stuff out into separate .css files too.

visualizer/index.html
```diff
@@ -455,6 +572,16 @@
       </div>
       <div id="parseResults"></div>
     </div>
+    <div id="actions">
```
pdubroy: Again, I think a more descriptive name here would be "semantics".

visualizer/ohmEditor.js
```diff
@@ -21,7 +21,13 @@ ohmEditor.registerEvents({
...
   // Emitted after attempting to parse the grammar and the input, respectively.
   'parse:grammar': ['matchResult', 'grammar', 'err'],
-  'parse:input': ['matchResult', 'trace']
+  'parse:input': ['matchResult', 'trace'],
+
+  // Emitted after changing the calling action
+  'change:action': ['trace'],
+
+  // Emitted after adding an new action
+  'add:action': ['actionType', 'actionName', 'optArguments']
```
pdubroy: Same as I mentioned at the top...are you sure this belongs on the ohmEditor?

visualizer/parseTree.js
```diff
@@ -645,6 +645,10 @@
     refreshParseTree(trace, true);
   });
 
+  ohmEditor.addListener('change:action', function(trace) {
```
pdubroy: I don't think you need this. Instead, just call `ohmEditor.parseTree.refresh()` from the semantics editor when the selected action is changed.

visualizer/actionButton.js
```diff
@@ -0,0 +1,265 @@
...
+      var actionName = actionNameContainer.value;
+      var args = actionType === 'Operation' ? retrieveActionArgs(actionWrapper) : undefined;
+      if (!actionName) {
+        actionContainer.removeChild(actionWrapper);
+        return;
+      } else if (argumentValueChangeOnly) {
+        // If the user is just changing an argument value, so we just
+        // refresh the tree with new arguments values
+        // TODO: update argument values
+        ohmEditor.emit('change:action', rootTrace);
```
pdubroy: So I think you can just call `ohmEditor.parseTree.refresh()` here instead.

pdubroy:
This looks pretty good! I think for now we should create a new CheckedEmitter named `ohmEditor.semantics`. You can do something similar to what we did for the parseTree: https://github.com/cdglabs/ohm/blob/master/visualizer/parseTree.js#L651. Then you can stick the event that's currently called `'change:action'` on there (I'd rename it to `'change:operation'`.)

I suggested a few other small changes, and after that feel free to merge.

lilyx:
Changes made correspondingly. I'll merge this for now so I could move forward. Let me know if anything needs to be modified.

pdubroy:
Looks good, thanks!

PR #107, opened 2016-05-23
Visualizer: add DOM for semantics editor

lilyx:


visualizer/actionButton.js
```diff
@@ -6,9 +6,29 @@
...
   } else {
-    initModule(root.ohmEditor);
+    initModule(root.ohmEditor, root.document);
   }
-})(this, function(ohmEditor) {
+})(this, function(ohmEditor, document) {
+  // DOM Helpers
+  // -----------
+
+  function createElement(sel, optContent) {
```
pdubroy: Let's not duplicate this code. I pushed this function (and a few other DOM helpers) into a new module: domUtil.js.

visualizer/actionButton.js
```diff
@@ -222,6 +232,7 @@
         // If the user is just changing an argument value, so we just
         // refresh the tree with new arguments values
         // TODO: update argument values
+        ohmEditor.semantics.emit('change:operationArgumentVals', args);
```
pdubroy: Nit: I'd just call this "operationArguments" -- it's kind of redundant to say "argument values".

src/pexprs-toArgumentNameList.js
```diff
@@ -56,7 +56,13 @@ pexprs.Terminal.prototype.toArgumentNameList = function(firstArgIndex) {
 };
 
 pexprs.Range.prototype.toArgumentNameList = function(firstArgIndex) {
-  return [this.from + '_to_' + this.to];
+  var argName = this.from + '_to_' + this.to;
+  // If the `argName` doesn't start with an valid character, i.e. alphabet or `_`,
+  // then prepending a `_` to it.
```
pdubroy: Grammar nits:
- "a valid", not "an valid" 
- "letter", not "alphabet"
- "prepend", not "prepending"

src/pexprs-toArgumentNameList.js
```diff
@@ -56,7 +56,13 @@ pexprs.Terminal.prototype.toArgumentNameList = function(firstArgIndex) {
...
+  if (!/[_$a-zA-Z]/.test(argName[0])) {
+    argName = '_' + argName;
```
pdubroy: This could still result in invalid identifiers, right? E.g. a range `"!".."@"`. I think the only time appending an underscore will help is if argName[0] is a digit. Also, this will incorrectly append an underscore for things that begin with non-ASCII characters (e.g. `ä`).

What you might want to do is write a utility function that will check if a given string is a valid JS identifier, and then in this method, first try `this.from + '_to_' + this.to`, then try appending `_`,  then fall back to a positional argument name (like `Terminal` does).

visualizer/index.html
```diff
@@ -128,37 +129,37 @@
...
       content: '\2014';
       padding: 0 0.5em;
     }
-    .pexpr > .label .caseName {
+    .pexpr > .self > .label .caseName {
       font-style: italic;
       opacity: 0.7;
     }
-    .pexpr:not(.failed) > .label:not(.prim):hover {
+    .pexpr:not(.failed) > .self:not(.nextStep) > .label:not(.prim):hover {
```
pdubroy: Is there a way you can do this in the semantics editor CSS instead of here? Since it's specific to the semantics editor, it doesn't really belong here.

visualizer/parseTree.js
```diff
@@ -436,11 +436,15 @@
...
     }
 
-    var label = wrapper.appendChild(createTraceLabel(traceNode));
+    var selfWrapper = wrapper.appendChild(createElement('.self'));
+    var label = selfWrapper.appendChild(createTraceLabel(traceNode));
 
     label.addEventListener('click', function(e) {
       if (e.altKey && !(e.shiftKey || e.metaKey)) {
         console.log(traceNode);  // eslint-disable-line no-console
+      } else if (e.metaKey && !e.shiftKey) {
```
pdubroy: Shouldn't we be checking that the alt key is not pressed here? E.g., `e.metaKey && !(e.altKey || e.shiftKey)`

visualizer/parseTree.js
```diff
@@ -436,11 +436,15 @@
...
+        // cmd + click to open or close semantic editor
+        ohmEditor.parseTree.emit('toggle:semanticsEditor', selfWrapper);
```
pdubroy: Let's call this 'cmdclick:traceElement' instead, and pass `wrapper` rather than `selfWrapper`.

visualizer/semanticsEditor.js
```diff
@@ -0,0 +1,336 @@
...
+'use strict';
+
+(function(root, initModule) {
+  if (typeof exports === 'object') {
+    module.exports = initModule;
+  } else {
+    initModule(root.ohm, root.ohmEditor, root.document, root.CodeMirror, root.CheckedEmitter);
+  }
+})(this, function(ohm, ohmEditor, document, CodeMirror, CheckedEmitter) {
+  // DOM Helpers
```
pdubroy: You can use the new domUtil.js here too.

visualizer/semanticsEditor.js
```diff
@@ -0,0 +1,336 @@
...
+
+  // Privates
+  // --------
+
+  // Creates a single `resultBlock`, which contains,
+  // `value`: the actual result container
+  // `operation`: the semantic operation for the result
+  function createResultBlock(opName, resultWrapper) {
+    var block = createElement('resultBlock');
+    block.classList.toggle('error', resultWrapper.isError);
```
pdubroy: You can use domUtil.toggleClasses here.

visualizer/semanticsEditor.js
```diff
@@ -0,0 +1,336 @@
...
+      editorWrapper.hidden = true;
+    } else {
+      editorWrapper.classList.add('resultOnly');
+    }
+  }
+  ohmEditor.parseTree.addListener('create:traceElement', function(wrapper, rootTrace, traceNode) {
+    var shouldHaveSemanticsEditor = ohmEditor.semantics.appendEditor &&
+        !wrapper.classList.contains('hidden') &&
+        !wrapper.classList.contains('failed') &&
+        !traceNode.expr.rulename !== 'spaces';
```
pdubroy: Why can't you write semantic actions for 'spaces'?

visualizer/semanticsEditor.js
```diff
@@ -0,0 +1,336 @@
...
+  // Creates a single operation argument tag, which contains a argument name, and
+  // corresponding value.
+  function createAndLoadArgTag(argName, argValue) {
+    var argTag = createElement('.argTag');
+    argTag.innerHTML = argName;
+
+    var valueSpan = argTag.appendChild(createElement('span'));
+    valueSpan.innerHTML = JSON.stringify(argValue);
+
+    // If the value is hided, hover the tag to temporarily show the argument value
```
pdubroy: Nit: "hidden", not "hided".

visualizer/semanticsHelper.js
```diff
@@ -6,26 +6,173 @@
...
   ohmEditor.addListener('parse:grammar', function(matchResult, grammar, error) {
     if (grammar && grammar.defaultStartRule) {
       semantics = grammar.semantics();
     }
   });
 
+  var opName = null;
+  var opArguments = null;
+  ohmEditor.semantics.addListener('change:semanticOperation', function(targetName, optArgs) {
+    if (opName === targetName) {
```
pdubroy: It's not obvious what this is doing -- why do you set `opName` to null here? Can you maybe add a comment explaining what this is doing?

visualizer/semanticsHelper.js
```diff
@@ -6,26 +6,173 @@
...
+      opName = null;
+    } else {
+      opName = targetName;
+    }
+    opArguments = ohmEditor.semantics.opArguments = optArgs;
+  });
+
+  ohmEditor.semantics.addListener('change:operationArgumentVals', function(optArgs) {
```
pdubroy: Do you need a separate event for this? Could you just use 'change:semanticOperation', since that is already passing the arguments?

visualizer/semanticsHelper.js
```diff
@@ -35,7 +182,99 @@
...
+    var key = nodeKey(traceNode.bindings[0]);
+    if (!(key in resultMap)) {
+      forceResult(traceNode);
+    }
+    return resultMap[key];
+  };
+
+  // If there is no user added action for the rule, return default argument list,
+  // Otherwise, return the argument list that user renamed before
+  ohmEditor.semantics.getActionArgPariedList = function(traceNode) {
```
pdubroy: Typo: should be "Paired", not "Paried"

visualizer/semanticsHelper.js
```diff
@@ -35,7 +182,99 @@
...
+    var defaultArgExpression = getDefaultArgExpression(traceNode);
+
+    var argPairList = {argExpr: defaultArgExpression};
+    var action = semantics._getActionDict(opName)[actionKey];
+    if (!action) {
+      return argPairList;
+    }
+
+    var actionStr = action.toString();
+    var isDefault = actionStr.indexOf('/* _DEFAULT_ */') + 15 === actionStr.indexOf('\n');
```
pdubroy: This is very fragile -- it's not obvious from the functions above that the comment `/* _DEFAULT_ */` is there for a reason. And in general, it's not a good idea to use string comparison to determine function identity -- there's too many subtle things that can go wrong.

What about just putting a known property on those functions? E.g., in `initActionDict`, you could do something like this:

```
var defaults = {
  _terminal: function() { ... },
  _iter: function(children) { ... },
  _nonterminal: function(children) { ... },
};
defaults.forEach(function(name, fn) {
  fn._isDefault = true;
});
```

Then here you could just check `action._isDefault`.

visualizer/style/semanticsEditor.css
```diff
@@ -0,0 +1,213 @@
+.pexpr:not(.failed) > .self.nextStep .label{
+  background: rgba(100,149,237,0.4);
```
pdubroy: Nit: Should have spaces after the commas.

lilyx:
Changes made, @pdubroy mind to take a look at again?

visualizer/semanticsEditor.js
```diff
@@ -0,0 +1,323 @@
...
+'use strict';
+
+(function(root, initModule) {
+  if (typeof exports === 'object') {
+    module.exports = initModule;
+  } else {
+    initModule(root.ohm, root.ohmEditor, root.document, root.CodeMirror, root.CheckedEmitter,
+        root.domUtil);
+  }
+})(this, function(ohm, ohmEditor, document, CodeMirror, CheckedEmitter, domUtil) {
```
pdubroy: Doesn't look like `document` is used anywhere now.

visualizer/index.html
```diff
@@ -539,8 +540,9 @@
   <script src="ohmEditor.js"></script>
   <script src="cmUtil.js"></script>
   <script src="domUtil.js"></script>
-  <script src="semanticsHelper.js"></script>
   <script src="parseTree.js"></script>
+  <script src="semanticseditor.js"></script>
```
pdubroy: Nit: For consistency, the filename should be camelcase ("semanticsEditor.js").

src/common.js
```diff
@@ -191,3 +191,8 @@ exports.unexpectedObjToString = function(obj) {
     return baseToString;
   }
 };
+
+exports.isValidJSIdentifier = function(str) {
+  var es5Grammer = require('../examples/ecmascript/es5').grammar;
```
pdubroy: Nit: "grammar".

src/common.js
```diff
@@ -191,3 +191,8 @@ exports.unexpectedObjToString = function(obj) {
     return baseToString;
   }
 };
+
+exports.isValidJSIdentifier = function(str) {
```
pdubroy: Hmmm, this is going to be pretty expensive to pull in the ES5 grammar and parse the string every time we want to check.

Now that I think about it more, I think it might be best to move this function into `toArgumentNameList`, and name it 'isRestrictedJSIdentifier' -- and then only allow ASCII characters (and just check with a simple regex). Otherwise there is a risk that we generate an identifier that the programmer doesn't actually know how to type!

visualizer/semanticsHelper.js
```diff
@@ -6,26 +6,181 @@
...
-    // Emitted after adding an new operation/attribute
-    'add:semanticOperation': ['type', 'name', 'optArguments']
-  });
+})(this, function(ohm, ohmEditor, CheckedEmitter) {
 
-  // Privates
-  // --------
   var semantics = null;
+
+  // `opName` is the semantic operation that current executing, it's null when every
```
pdubroy: Nit: "whenever", not "when every"

visualizer/semanticsHelper.js
```diff
@@ -6,26 +6,181 @@
...
   ohmEditor.addListener('parse:grammar', function(matchResult, grammar, error) {
     if (grammar && grammar.defaultStartRule) {
       semantics = grammar.semantics();
+      opName = null;
+      opArguments = null;
     }
   });
 
+  ohmEditor.semantics.addListener('change:semanticOperation', function(targetName, optArgs) {
+    if (opName === targetName) {
```
pdubroy: I still don't understand why you clear opName if it's equal to targetName...seems a bit odd, and worth a comment explaining why.

pdubroy:
Thanks! LGTM if you make the changes I suggested.

PR #108, opened 2016-05-25
Visualizer: let semantics editor handle loading and saving

lilyx:


visualizer/parseTree.js
```diff
@@ -609,15 +609,15 @@
           visibleChoice: isVisibleChoice(node, parent)
         });
 
+        ohmEditor.parseTree.emit('create:traceElement', el, rootTrace, node);
```
pdubroy: I moved this below the `isCollapsed` stuff because it seems to me that we haven't really finished creating the trace element here. I think this event is more generally useful if the element is really "done" when we call it.

Is there a reason you need it be here instead?

visualizer/parseTree.js
```diff
@@ -717,8 +718,10 @@
   // -------
 
   var parseTree = ohmEditor.parseTree = new CheckedEmitter();
-  parseTree.refresh = refreshParseTree;
-
+  parseTree.refresh = function(rootTrace, clearZoomState) {
+    clearMarks();
```
pdubroy: I think maybe we should just be clearing the marks inside `refreshParseTree`. What do you think?

visualizer/semanticsEditor.js
```diff
@@ -45,25 +48,34 @@
...
+          resultBlock.classList.toggle('leftBorder', idx++ > 0);
+        }
+        if (!resultWrapper.forced && resultWrapper.isNextStep) {
           selfWrapper.classList.add('nextStep');
         }
-        selfWrapper.classList.toggle('passThrough', resultWrapper.isPassThrough);
+
         if (resultWrapper.forCallingSemantic) {
           selfWrapper._args = resultWrapper.args;
+          selfWrapper.classList.toggle('passThrough', !!resultWrapper.isPassThrough);
```
pdubroy: Nit: do you really need the coercion (`!!`) here? `isPassthrough` already implies that it's a boolean -- so if necessary, you should probably do the coercion when you set `isPassthrough`.

visualizer/semanticsEditor.js
```diff
@@ -317,6 +378,8 @@
     'add:semanticOperation': ['type', 'name', 'optArguments'],
 
     // Emitted after changing to another semantic operation
-    'change:semanticOperation': ['targetName', 'optArguments']
+    'change:semanticOperation': ['targetName', 'optArguments'],
+
+    'save:action': ['traceNode', 'actionArguments', 'actionBody']
```
pdubroy: I'd name this 'save:semanticAction' for consistency.

visualizer/semanticsHelper.js
```diff
@@ -259,12 +287,81 @@
     return pexpr;
   }
 
+  // Build the action that called from the wrapper, which includes renaming the
+  // operation arguments, and actual action body that user typed.
+  function buildAction(opArgs, enclosedActionArgStr, actionBody) {
+    // Constructs the codes that rename the operation arguments, so we could refer
```
pdubroy: Nit: Should be "the code that renames", not "the codes that rename"

visualizer/semanticsHelper.js
```diff
@@ -259,12 +287,81 @@
...
+        var action = buildAction(this.args, enclosedActionArgStr, actionBody);
+        result = action.apply(this, arguments);
+        if (todoList) {
+          result = failure;
+        }
+      } catch (error) {
+        if (todoList) {
+          result = failure;
+        } else {
+          result = error.isErrorWrapper ? error : new ErrorWrapper(key, error);
```
pdubroy: I don't think there's any point reassigning `result` here -- you can just do `throw error.isErrorWrapper ? error : new ErrorWrapper(key, error);`

visualizer/semanticsHelper.js
```diff
@@ -259,12 +287,81 @@
...
+    saveAction(traceNode, opName,  actionArguments, actionBody);
+  });
+
   // Exports
   // -------
 
   ohmEditor.semantics.getResults = function(traceNode) {
     var key = nodeKey(traceNode.bindings[0]);
-    if (!(key in resultMap)) {
+    // If the node has not been evaluated yet, or all its exist smantics results are
```
pdubroy: "all its existing semantic results"

visualizer/parseTree.js
```diff
@@ -662,6 +662,7 @@
   // Refresh the parse tree after attempting to parse the input.
   ohmEditor.addListener('parse:input', function(matchResult, trace) {
     $('#bottomSection .overlay').style.width = 0;  // Hide the overlay.
+    ohmEditor.parseTree.rootTrace = trace;
```
pdubroy: Does the root trace need to be public? It seems like this only references to it are to pass it back in as an argument to `ohmEditor.parseTree.refresh`. If that's the case, then we can probably just get rid of that argument, and keep the root trace in a variable inside of parseTree.js.

pdubroy: I just pushed a change that refactors parseTree.js a bit to avoid passing around the rootTrace everywhere -- hopefully this still works you, let me know if you have any problems.

visualizer/parseTree.js
lilyx: I just need to check the result to found out whether there is a next step node that will be hided by collapsing, and the result is coming from the listener of `create:traceElement`. Anyway, I moved it back, and let the `create:traceElement` listener check the collapse status instead.

visualizer/parseTree.js
lilyx: Sounds good to me. I try to move it, and the test is complaining about cannot find `cmUtil.clearMark`, not sure how you want to handle this, so I just keep it here for now.

visualizer/semanticsHelper.js
lilyx: There is a `finally` block that records the `result` to resultMap, and we want to record the `error` here.

lilyx:
Changes made. @pdubroy let me know if there is anything needs to be modified.

visualizer/semanticsHelper.js
```diff
@@ -259,12 +289,82 @@
...
+    // Constructs the code that renames the operation arguments, so we could refer
+    // them by name directly. i.e. `var <argument name> = this.args.<argument name>`
+    var renameOpArgStr = opArgs &&  Object.keys(opArgs).map(function(argName) {
+      return 'var ' + argName + ' = this.args.' + argName + ';';
+    }).join('\n');
+
+    var actionStr = 'function' + enclosedActionArgStr + '{\n' +
+        renameOpArgStr + '\n' + actionBody +
+      '\n}';
+    return eval('(' + actionStr + ')');    // eslint-disable-line no-eval
```
pdubroy: I think you should actually use the Function constructor here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function. It's basically like `eval`, but it's _slightly_ safer, because the code doesn't have access to local scope variables (though it still has access to the global object).

visualizer/semanticsHelper.js
pdubroy: Ok, this control flow is *really* tricky to understand. In generally you should use `finally` with caution, because it's really hard to understand to follow the flow of code like this. Can you rewrite it in a way that doesn't use a finally block? I'll leave a few more comments with suggestions.

visualizer/semanticsHelper.js
```diff
@@ -259,12 +289,82 @@
...
+    if (isExpression) {
+      actionBody = 'return ' + actionBody + ';';
+    }
+    var wrapper = function() {
+      var key = nodeKey(this);
+      var result;
+      var origTodoList = todoList;
+      todoList = undefined;
+      try {
+        var action = buildAction(this.args, enclosedActionArgStr, actionBody);
```
pdubroy: It looks like this line could throw due to the `eval` in `buildAction`, right? Then I would put this line in it's own try/catch block and handle this error separately from other kinds of error.

visualizer/semanticsHelper.js
```diff
@@ -259,12 +289,82 @@
...
+        result = action.apply(this, arguments);
```
pdubroy: ...and then I would put this line in a separate try/catch as well, because there is a completely different type of error that might get thrown here. Then I think you might be able to get rid of the `isErrorWrapper` check below.

visualizer/semanticsHelper.js
```diff
@@ -259,12 +289,82 @@
...
+        if (todoList) {
+          result = failure;
+        }
+      } catch (error) {
+        if (todoList) {
```
pdubroy: It would be nice if you didn't have to duplicate this check.

visualizer/semanticsHelper.js
```diff
@@ -259,12 +289,82 @@
...
+          result = failure;
+        } else {
+          result = error.isErrorWrapper ? error : new ErrorWrapper(key, error);
+          throw result;
```
pdubroy: Can you avoid throwing here, and instead return an error value from this function? I think that would help simplify the control flow. Maybe `wrapAction` could either return a function (on success), or `null` (on failure).

pdubroy:
Hi Meixian, I think it would be a good idea to clean up the control flow logic in the try/catch/finally block, as I've suggested in my comments above. Otherwise everything looks good. Thanks!

lilyx:
I get ride of the `finally` block, and add some comments, hope this helps. As there are still two more pull requests left, so I'll go ahead and merge this part. Let me know if it's still unclear.

pdubroy:
Great! Looks much easier to follow now. Also I saw you cleaned up the other try/catch/finally blocks as well -- thanks 😄

PR #109, opened 2016-05-25
JSON recipes for grammars

mroeder:
Implementation of JSON recipes for grammars. Semantics are still JavaScript (functions) since they are also implemented in JavaScript for Ohm/JS.

Referencing #70 here (it's not a `toJSON` but switching over `toRecipe`).


src/Builder.js
```diff
@@ -14,10 +14,43 @@ var pexprs = require('./pexprs');
...
+      gDecl.withSource(metaInfo.source);
+    }
+
+    var self = this;
+    this.currentDecl = gDecl;
+    Object.keys(rules).forEach(function(ruleName) {
+      var ruleRecipe = rules[ruleName];
+
+      var action = ruleRecipe[0]; // define/extend/override
+      // var metaInfo = ruleRecipe[1]; // TODO: us and handle meta-info
```
mroeder: Rules currently don't have meta-info attached. We could leave this out (and simplify the call below) or just go with the general format that `[1]` is the meta-info in JSON recipes and therefore keep it empty.
@pdubroy @alexwarth

src/Grammar.js
```diff
@@ -146,52 +146,70 @@ Grammar.prototype = {
...
-        sb.append('define');
+        rule.push('define');
       }
-      var formals = self.ruleFormals[ruleName];
-      var formalsString = '[' + formals.map(JSON.stringify).join(', ') + ']';
-      sb.append('(' + JSON.stringify(ruleName) + ', ' + formalsString + ', ');
-      body.outputRecipe(sb, formals, self.definitionInterval);
+
+      // TODO: add meta-info
+      rule.push({});
```
mroeder: (see above in the Builder comment)

src/pexprs-outputRecipe.js
```diff
@@ -36,91 +28,95 @@ function getIntervalInfo(expr, grammarInterval) {
...
+pexprs.Apply.prototype.outputRecipe = function(formals, grammarInterval) {
+  return [
+    'app',
+    getMetaInfo(this, grammarInterval),
+    this.ruleName,
+    this.args.map(function(arg) {
+      return arg.outputRecipe(formals, grammarInterval);
+    })
+  ];
+  // TODO: fold this into the main recipe creation (since _ indicates case)
```
mroeder: What's currently still missing, is that the rule-level alteration,
```
rule = sub1 -- one
     | sub2 -- two
```
gets folded into the rule definition. Just as with the old recipe format, those inline rules get serialized separately:
```json
{
  "rule_one": ["define",{},null,[],["app",{},"sub1",[]]],
  "rule_two": ["define",{},null,[],["app",{},"sub2",[]]],
  "rule": ["define",{},null,[],
    ["alt",{},
      ["app",{},"rule_one",[]],
      ["app",{},"rule_two",[]]
    ]
  ]
}
```
It would be possible to fold that to something like:
```json
{
  "rule": ["define",{},null,[],
    ["app",{"case":"one"},"sub1",[]],
    ["app",{"case":"two"},"sub2",[]]
  ]
}
```
Preferences? Opinions? @alexwarth @pdubroy

src/Grammar.js
```diff
@@ -146,52 +146,70 @@ Grammar.prototype = {
...
-    // Include the supergrammar in the recipe if it's not a built-in grammar.
-    var superGrammarDecl = '';
-    if (!this.superGrammar.isBuiltIn()) {
-      sb.append(this.superGrammar.toRecipe('buildSuperGrammar'));
-      superGrammarDecl = '    .withSuperGrammar(buildSuperGrammar.call(this))\n';
-    }
-    sb.append('  var decl = this.newGrammar(' + JSON.stringify(this.name) + ')\n');
+    var recipe = [
+      'grammar',
+      {}, // meta-info
```
pdubroy: I'd stick this in a var named `metaInfo`, then you can eliminate the comment, and the code below that uses `recipe[1]` will be a bit more self-documenting.

src/pexprs-outputRecipe.js
```diff
@@ -36,91 +28,95 @@ function getIntervalInfo(expr, grammarInterval) {
...
+  ];
 };
 
-pexprs.Lookahead.prototype.outputRecipe = function(sb, formals, grammarInterval) {
-  sb.append('this.la(');
-  this.expr.outputRecipe(sb, formals, grammarInterval);
-  sb.append(')' + getIntervalInfo(this, grammarInterval));
+pexprs.Lookahead.prototype.outputRecipe = function(formals, grammarInterval) {
+  return [
+    'la',
```
pdubroy: While we're at it, why don't we change this to `lookahead`? I don't think we're getting a big savings here :-)

src/pexprs-outputRecipe.js
pdubroy: I don't see a big advantage to folding them in. They are the same as top-level rules for all intents and purposes, so I think it's fine for them to be like that in the recipe.

src/Builder.js
pdubroy: One thing I need for the visualizer, but hadn't implemented yet, is the "definition interval" of a rule. E.g., given a grammar `G { foo = digit }`, the definitionInterval would be `[4, 15]`. So it would definitely be useful to be able to have rule metadata.

src/Builder.js
```diff
@@ -34,6 +67,9 @@ Builder.prototype = {
     var terms = [];
     for (var idx = 0; idx < arguments.length; idx++) {
       var arg = arguments[idx];
+      if (!(arg instanceof pexprs.PExpr)) {
```
pdubroy: It would be nice to avoid having to do this in every one of the constructor functions. Is there a reason that you can't do this from the bottom up, by calling `toRecipe` on all the arguments before invoking the constructor function?

src/main.js
```diff
@@ -332,8 +332,16 @@ function grammarsFromScriptElements(optNodeOrNodeList) {
   return ns;
 }
 
-function makeRecipe(recipeFn) {
-  return recipeFn.call(new Builder());
+function makeRecipe(recipe) {
+  if (typeof recipe === 'function') {
+    return recipe.call(new Builder());
+  } else {
```
pdubroy: Is it valid to call this with an argument that is neither a function nor a string? If not, consider making this an `else if` and add a `else` case that throws an error.

src/pexprs-outputRecipe.js
pdubroy: Also note that we still allow `_` in regular rule names, so you can't detect easily detect inline rules just based on the name.

src/pexprs-outputRecipe.js
mroeder: Good idea! Will do.

src/main.js
mroeder: If it is an object, it would count as a deserialized JSON recipe. But maybe it's cleaner to only allow "stringified" JSON recipes with `makeRecipe`.

src/Builder.js
mroeder: Only the constructor function knows what the arguments are. Since not every argument is `PExpr` but some are just plain values, `fromRecipe` cannot do a reconstruction.
What would it reconstruct anyway? Builder methods call other builder methods to reconstruct their arguments/sub-PExprs.

src/Builder.js
```diff
@@ -57,31 +99,76 @@ Builder.prototype = {
...
+        return param instanceof pexprs.PExpr ? param :
+          this.fromRecipe(param);
+      }, this);
+    }
     return new pexprs.Apply(ruleName, optParams);
+  },
+
+  fromRecipe: function(recipe) {
+    // the meta-info of 'grammar' is proccessed in Builder.grammar
+    var res = this[recipe[0]].apply(this,
```
pdubroy: Is `res` should for `result` here? It took me a while to figure out -- I kept thinking it meant "resource" or something. It's a minor point, but I would either call this "result" or "ans", which are the two most common names we use elsewhere in Ohm.

src/Builder.js
pdubroy: I was thinking that everything of the form [ctorName, ...] could be recursively traversed, and constructed in a bottom-up manner. E.g.:

```js
function fromRecipe(r) {
  if (!Array.isArray(r)) {
    return r;  // Strings, Numbers, etc. don't need further processing.
  }
  var ctorName = r[0];
  var args = r.slice(2);
  var result = ctorName.apply(this, args.map(fromRecipe));
  ...
  return result;
}
```

This way, the individual constructors (`plus`, `opt`, `star`, etc.) wouldn't have to call `fromRecipe`. Of course this wouldn't work for cases where a constructor takes an Array as an argument. And maybe it doesn't work for other reasons -- you've thought about this more. Just wanted to clarify what I was trying to suggest.

src/main.js
pdubroy: Up to you. I'm not sure if there's a good reason to accept deserialized recipes or not.

src/pexprs-outputRecipe.js
```diff
@@ -36,91 +26,96 @@ function getIntervalInfo(expr, grammarInterval) {
 
 pexprs.PExpr.prototype.outputRecipe = common.abstract('outputRecipe');
 
-pexprs.any.outputRecipe = function(sb, formals, grammarInterval) {
-  throw new Error('should never output a recipe for `any` expression');
+pexprs.any.outputRecipe = function(formals, grammarInterval) {
+  return [
```
pdubroy: Nit: I would put this all on one line.

src/pexprs-outputRecipe.js
```diff
@@ -36,91 +26,96 @@ function getIntervalInfo(expr, grammarInterval) {
...
+    'any',
+    getMetaInfo(this, grammarInterval)
+  ];
 };
 
-pexprs.end.outputRecipe = function(sb, formals, grammarInterval) {
-  throw new Error('should never output a recipe for `end` expression');
+pexprs.end.outputRecipe = function(formals, grammarInterval) {
+  return [
```
pdubroy: Ditto.

pdubroy:
LGTM, with a few suggestions.

src/main.js
mroeder: I will leave it in. Currently, semantics recipes and the bootstrapped JS-files (dist/built-in-rules.js, etc.) use "JSO"(N) with `makeRecipe`. It's faster - since it does not have to be deserialized - and it also circumvents escaping quotes (lots of strings in strings if you put JSON in a string first).
For pure grammar recipes, this doesn't matter and they can be read and processed as strings.

PR #110, opened 2016-05-27
Improve es5 errors

tonyg:
If there's a rule description on Statement, it masks a number of interesting errors. I saw this when extending the ES5 grammar for my Syndicate/js DSL. I add new cases to Statement expecting input such as `on message foo() { ... }`, and if I forget the `message`, I get an unhelpful error far away from the real problem. Removing the rule description from Statement gets me much more useful error reports.

(Separately, I've fixed a typo: `indentifier` => `identifier`.)


alexwarth:
Thanks, Tony!

(Earlier today @tonyg and I talked about a couple of things that could be done about this, e.g., nesting failures. We'll think about this some more, and loop in @lilyx before we do anything.)

PR #111, opened 2016-06-01
Visualizer: add context menu to action buttons for `edit` and `delete`

lilyx:


visualizer/actionButton.js
```diff
@@ -279,4 +366,17 @@
     document.querySelector('#operations').innerHTML = '';
     document.querySelector('#attributes').innerHTML = '';
   });
+
+  // Hide the action context menu when Esc or Enter is pressed, any click happens, or another
+  // context menu is brought up.
+  document.addEventListener('click', hideContextMenu);
```
lilyx: I duplicated the code here to hide the context menu for operation button, maybe we should put the hiding code together in a separate file?

src/Semantics.js
```diff
@@ -413,7 +413,7 @@ Semantics.prototype.addOperationOrAttribute = function(type, signature, actionDi
       return '[' + name + ' operation]';
     };
   } else {
-    Object.defineProperty(this.Wrapper.prototype, name, {get: doIt});
+    Object.defineProperty(this.Wrapper.prototype, name, {get: doIt, configurable: true});
```
pdubroy: Nit: Might be nice to break this across lines, and add a comment why it needs to be configurable. E.g.:
```js
Object.defineProperty(this.Wrapper.prototype, name, {
  get: doIt,
  configurable: true  // So the property can be deleted.
});
```

visualizer/actionButton.js
```diff
@@ -12,6 +12,10 @@
 
   // Privates
   // --------
+  var KeyCode = {
```
pdubroy: Maybe stick this in domUtil, so it can be shared by all code that needs it? It's pretty useful.

visualizer/actionButton.js
```diff
@@ -20,15 +24,146 @@
...
+    var argList = wrapper.querySelector('.arguments');
+    var nameValPairs = wrapper.querySelectorAll('.nameValuePair');
+    var args = Object.create(null);
+    Array.prototype.forEach.call(nameValPairs, function(nameValPair) {
+      var argNameContainer = nameValPair.querySelector('.name');
+      var argName = argNameContainer.value;
+
+      var valueContainer = nameValPair.querySelector('.value');
+      var value = valueContainer.value;
+      if (argName && /^[_a-zA-Z0-9]+$/.test(argName) && !(argName in args)) {
```
pdubroy: What's this regex doing? In the very least this could use a comment. It looks like it's testing for valid JavaScript identifiers, but we should be very careful doing that if the user has specified the name of the identifier. The other case was a bit different, because we are _generating_ the identifier name, so I think it's ok to restrict that to ASCII characters. But it looks like this is taking a name that comes from the user, and testing if it looks like a valid identifier -- is that right?

If that's the case, we should at least add a TODO here that says we should fix this to allow any valid JavaScript identifier, not just the restricted set that will match this regex.

visualizer/actionButton.js
```diff
@@ -20,15 +24,146 @@
...
+        args[argName] = eval(value);  // eslint-disable-line no-eval
```
pdubroy: So, funny things happen if you use `args` as the value of an argument :-)

As I mentioned in the last PR, it's usually slightly better to use `new Function` rather than `eval`, because `eval` gives access to all the variables in the scope where it is called.

visualizer/actionButton.js
```diff
@@ -154,6 +253,24 @@
     buttonWrapper.hidden = false;
   }
 
+  function handleContextMenuOnAction(nameContainer) {
```
pdubroy: Nice! All your DOM handling code is very nice and clear, good names and useful comments.

visualizer/actionButton.js
```diff
@@ -223,44 +332,22 @@
...
-          wrapper.querySelector('.buttonWrapper').querySelector('.add').hidden = true;
-          wrapper.querySelector('.buttonWrapper').querySelector('.arrow').hidden = false;
+          wrapper.querySelector('.buttonWrapper .add').hidden = true;
+          wrapper.querySelector('.buttonWrapper .arrow').hidden = false;
         }
       }
       // TODO: add operation/attribute to force evaluation menu
       try {
-        ohmEditor.semantics.emit('add:semanticOperation', type, name, args);
+        var origActionDict = wrapper._origActionDict;
```
pdubroy: It would be good to move everything that can't throw an error outside of the `try` block. That way it's very clear where the error can come from (and gives a better idea of what kind of errors are possible).

visualizer/actionButton.js
pdubroy: Good idea. I think it's fine to merge this as-is, and clean up the duplication in a separate PR.

How about we move the contextMenu stuff to domUtil?

visualizer/index.html
```diff
@@ -372,8 +372,13 @@
     <div class="externalRules"><div class="content"></div></div>
   </div>
 
-  <div id="contextMenu" hidden><ul></ul></div>
-
+  <div id="node" class="contextMenu" hidden><ul></ul></div>
```
pdubroy: Ideally, the `id` should make sense on its own, without knowing the class name. Here, "node" seems like much too generic of a name. What about "parseTreeMenu"?

visualizer/index.html
```diff
@@ -372,8 +372,13 @@
...
+  <div id="action" class="contextMenu" hidden>
```
pdubroy: ...and then I would call this "operationMenu".

visualizer/semanticsHelper.js
```diff
@@ -201,21 +201,44 @@
     return defaults;
   }
 
+  // Merge the actions in the `optActionDict` into `actionDict`
+  function mergeActionDict(name, actionDict, optActionDict) {
+    if (!optActionDict) {
```
pdubroy: Do you need it both of this `if` conditions?

visualizer/semanticsHelper.js
```diff
@@ -201,21 +201,44 @@
...
       }
     }
+
     semantics['add' + type](signature, initActionDict(type, name));
+    mergeActionDict(name, semantics._getActionDict(name), optOrigActionDict);
   }
-  ohmEditor.semantics.addListener('add:semanticOperation', function(type, name, optArguments) {
+  ohmEditor.semantics.addListener('add:semanticOperation', function(type, name, optArguments,
+      origActionDict) {
+    if (!/^[a-zA-Z_$][0-9a-zA-Z_$]*$/.test(name)) {
```
pdubroy: Again, I would factor this check out into a separate function (e.g. `isOperationNameValid`), and a comment that explains what it should be checking (and why), and maybe a TODO to make it less restrictive in the future.

pdubroy:
Mostly looks good, feel free to merge after you make the changes I suggested.

visualizer/actionButton.js
lilyx: Sounds good to me! `domUtil` is a reasonable place to put this part, and the `KeyCode`.

visualizer/actionButton.js
lilyx: It's a bug. The regex should test whether the name is a restrict JS identifier. Fixed it.

PR #112, opened 2016-06-03
Visualizer: add `Force Evaluation` to `parseTreeMenu`

lilyx:


visualizer/index.html
```diff
@@ -253,6 +253,7 @@
       padding: 0;
     }
     .contextMenu li {
+      position: relative;
```
pdubroy: What's this for?

visualizer/semanticsEditor.js
```diff
@@ -360,6 +366,177 @@
   }
   ohmEditor.parseTree.addListener('cmdclick:traceElement', toggleSemanticsEditor);
 
+  // Retrieve arguments from the menu entry's submenu
+  function retrieveArgumentsFromSubList(entry) {
+    var argumentWrappers = entry.querySelectorAll('ul li');
+    if (!argumentWrappers) {
+      return undefined;
```
pdubroy: Nit: You can just do `return;` to return `undefined`.

visualizer/semanticsEditor.js
```diff
@@ -360,6 +366,177 @@
...
+      var value = eval(valueContainer.value);    // eslint-disable-line no-eval
+      operationArguments[name] = value;
+    });
+    return operationArguments;
+  }
+
+  // Create a menu entry that wraps the argument name and a textarea for user to enter
+  // corresponding value.
+  function createArgumentWrapper(name) {
+    var argumentWrapper = domUtil.createElement('li');
```
pdubroy: I think this code would be slightly easier to read if you use a shorter name for this variable. Since we're already in a function called `createArgumentWrapper`, you could probably just call it `wrapper`.

visualizer/semanticsEditor.js
```diff
@@ -360,6 +366,177 @@
...
+    argumentWrapper.appendChild(domUtil.createElement('.name', name));
+    argumentWrapper.appendChild(domUtil.createElement('.assign', UnicodeChars.LEFTWARDS_ARROW));
+    var valueContainer = argumentWrapper.appendChild(domUtil.createElement('textarea.value'));
+    valueContainer.placeholder = 'value';
+    valueContainer.cols = 5;
+    valueContainer.addEventListener('keyup', function(e) {
```
pdubroy: I like to do `onkeyup = function(e) { ... }` if I know that there is only one listener. That way, you know that if someone needs to add a second listener for the same event, they will look at this code (and possibly combine the two).

visualizer/semanticsEditor.js
```diff
@@ -360,6 +366,177 @@
...
+      valueContainer.cols = Math.max(valueContainer.value.length, 5);
+    });
+    valueContainer.addEventListener('click', function(e) {
+      e.stopPropagation();
+    });
+    valueContainer.addEventListener('keypress', function(e) {
```
pdubroy: Same with these two, I'd do `onclick = ...` and `onkeypress = ...`.

visualizer/semanticsHelper.js
```diff
@@ -253,7 +277,7 @@
       if (operationName in semantics._getSemantics().operations) {
         var argValues;
         if (operationName === opName) {
-          argValues = toValueList(opArguments || Object.create(null));
+          argValues = toValueList(optArguments || opArguments || Object.create(null));
```
pdubroy: Consider renaming one of these variables -- `opArguments` and `optArguments` are *really* similar, and it's easy to get confused.

visualizer/semanticsHelper.js
```diff
@@ -419,13 +443,29 @@
   // Exports
   // -------
 
+  ohmEditor.semantics.getResult = function(traceNode, name, optArguments) {
+    var key = nodeKey(traceNode.bindings[0]);
+    forcing = true;
+    if (hasResult(key, name, optArguments)) {
+      forcing = false;
+      return undefined;
```
pdubroy: Nit: Just `return;` here.

visualizer/style/semanticsEditor.css
```diff
@@ -209,3 +217,32 @@ headerBlock real {
...
+}
+.contextMenu li.disabled ul {
+  display: none;
+}
+.contextMenu ul ul {
+  left: 100%;
+  top: 0;
+  padding: 0;
+  position: absolute;
+  border-radius: 4px;
```
pdubroy: Can you keep these in alphabetical order? I like to do it that way, since otherwise the order tends to get a bit random. It's fine that some CSS doesn't use alphabetical order, but let's keep the stuff that is already ordered consistent.

visualizer/style/semanticsEditor.css
```diff
@@ -209,3 +217,32 @@ headerBlock real {
   color: #999;
 }
 
+/* `Force Evaluation` entry in `parseTreeMenu` */
+.contextMenu label span {
```
pdubroy: This is too generic -- other labels may want to include a span, and they shouldn't automatically get a `margin-left`. Maybe add an appropriately-named class to the span.

visualizer/style/semanticsEditor.css
```diff
@@ -209,3 +217,32 @@ headerBlock real {
...
+  margin-left: 10px;
+}
+.contextMenu li.disabled ul {
```
pdubroy: Disabled items shouldn't be hidden, they should be greyed out. If you need to hide an item, just don't add it to the menu when you're handling the `contextMenu` event.

visualizer/style/semanticsEditor.css
```diff
@@ -209,3 +217,32 @@ headerBlock real {
...
+  display: none;
+}
+.contextMenu ul ul {
```
pdubroy: All of this CSS probably belongs with the other contextMenu CSS.

pdubroy:
Mostly looks good, feel free to commit after fixing these issues.

visualizer/index.html
lilyx: I added submenu to the `Force Evaluation`, this is used for aligning the submenu (in the `styl/semanticsEditor.css`)

visualizer/semanticsEditor.js
lilyx: Then style checking will throw an error said that the other `return` statement in this method shouldn't return anything. i.e. `383:4  error  Expected no return value  consistent-return`

visualizer/semanticsHelper.js
lilyx: The same as the other one, the style checking will complain.

visualizer/index.html
pdubroy: Ah, got it. Maybe leave a comment here then? Like `/* For position submenus */`.

visualizer/semanticsEditor.js
pdubroy: Ah, dang. Maybe we should think about turning that check off -- I'm not sure how useful it is. Don't worry about it for now though, just leave this code as-is.

PR #116, opened 2016-06-24
Allow ohm-cmd to generate recipe for ProtoBuiltInRules

tonyg:
Allow ohm-cmd to generate recipe for ProtoBuiltInRules; useful for bootstrapping. (Second PR; I messed the first one up)

### DIFF
--- a/src/Grammar.js
+++ b/src/Grammar.js
@@ -142,11 +142,6 @@ Grammar.prototype = {
   },
 
   toRecipe: function(optVarName) {
-    if (this.isBuiltIn()) {
-      throw new Error(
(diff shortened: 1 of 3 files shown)


src/Grammar.js
```diff
@@ -162,12 +162,13 @@ Grammar.prototype = {
     var self = this;
     Object.keys(this.ruleBodies).forEach(function(ruleName) {
       var body = self.ruleBodies[ruleName];
+      var isDefinition = !self.superGrammar || !self.superGrammar.ruleBodies[ruleName];
```
alexwarth: Even better, thanks.

PR #118, opened 2016-06-29
Semantics editor updates

lilyx:
Couple of corrections on result display for specified cases.


visualizer/semanticsEditor.js
```diff
@@ -66,6 +101,12 @@
       opSignature += '(' + argValues.join(',') + ')';
     }
     var opNameContainer = block.appendChild(domUtil.createElement('operation'));
+    var semanticOperations = ohmEditor.semantics.getSemantics();
```
pdubroy: It seems like this var should just be called `semantics`.

visualizer/semanticsEditor.js
```diff
@@ -66,6 +101,12 @@
...
+    var operationCounts = Object.keys(semanticOperations.operations).length +
```
pdubroy: Nit: Should probably be named "operationCount", not "operationCounts", since it's just one number.

visualizer/semanticsEditor.js
```diff
@@ -34,6 +34,41 @@
...
+      return;
+    }
+
+    var resultContainer = targetBlock.parentElement;
+    resultContainer.classList.add('showing');
+    Array.prototype.forEach.call(resultContainer.children, function(block) {
+      if (block !== targetBlock) {
+        block.classList.add('hidden');
+      } else {
+        targetBlock.classList.add('showing');
```
pdubroy: I think you could write this more succinctly as:

```js
block.classList.add(block === targetBlock ? 'showing' : 'hidden');
```

visualizer/semanticsEditor.js
```diff
@@ -34,6 +34,41 @@
...
+      }
+    });
+  }
+
+  // Remove the "markers" that used for display a single result block, in order
```
pdubroy: Grammar nit: "that are used to display..."

pdubroy:
LGTM, with a couple small suggestions.

PR #119, opened 2016-06-30
Visualizer: fix the problem of displaying `pexpr.Opt` as argument

lilyx:
This fixed the bug for displaying Iter expressions as arguments


pdubroy:
lgtm

PR #122, opened 2016-07-01
Visualizer: move `$` and `$$` to domUtil, and replace domcument queries.

lilyx:
This moves the definition code for `$` & `$$` into domUtil, and replace `document.querySelector` and `document.querySelectorAll`

### DIFF
--- a/visualizer/parseTree.js
+++ b/visualizer/parseTree.js
@@ -11,7 +11,7 @@
   }
 })(this, function(ohm, ohmEditor, CheckedEmitter, document, cmUtil, d3, domUtil) {
   var ArrayProto = Array.prototype;
-  function $(sel) { return document.querySelector(sel); }
+  var $ = domUtil.$;
 
   var UnicodeChars = {
     ANTICLOCKWISE_OPEN_CIRCLE_ARROW: '\u21BA',
--- a/visualizer/semanticsEditor.js
+++ b/visualizer/semanticsEditor.js
@@ -13,6 +13,8 @@
 
   // Privates
   // --------
+  var $ = domUtil.$;
+  var $$ = domUtil.$$;
 
   var UnicodeChars = {
     PLUS_SIGN: '\u002B',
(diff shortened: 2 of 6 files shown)


visualizer/parseTree.js
```diff
@@ -11,7 +11,6 @@
   }
 })(this, function(ohm, ohmEditor, CheckedEmitter, document, cmUtil, d3, domUtil) {
   var ArrayProto = Array.prototype;
-  function $(sel) { return document.querySelector(sel); }
```
pdubroy: Since we use `$` so much in this file, I'd be tempted to put the following at the top of the file:

```js
var $ = domUtil.$;
```

visualizer/semanticsEditor.js
```diff
@@ -107,13 +107,13 @@
     // Hover the block, and all the blocks that represent the results for the same operation
     // signature will be highlighted.
     block.onmouseover = function(event) {
-      var blocks = document.querySelectorAll('.semanticsEditor .result .' + blockClassId);
+      var blocks = domUtil.$$('.semanticsEditor .result .' + blockClassId);
       Array.prototype.forEach.call(blocks, function(b) {
```
pdubroy: Oh, you can just do `blocks.forEach` now!

visualizer/semanticsEditor.js
```diff
@@ -107,13 +107,13 @@
...
         b.classList.add('highlight');
       });
     };
     block.onmouseout = function(event) {
-      var blocks = document.querySelectorAll('.semanticsEditor .result .' + blockClassId);
+      var blocks = domUtil.$$('.semanticsEditor .result .' + blockClassId);
       Array.prototype.forEach.call(blocks, function(b) {
```
pdubroy: `blocks.forEach`

pdubroy:
lgtm with a few small suggestions. Thanks!

PR #124, opened 2016-07-06
Fixes for layout issues in visualization

sakekasi:
The current visualization has several small bugs in its layout, with long tokens and with arrows in showFailures mode. This pull request will address these issues


pdubroy:
Cool, this seems to make things better, but it's still not quite right in some cases:

screen shot 2016-07-06 at 10 54 36 am

sakekasi:
FYI, this PR seeks to address #128

visualizer/parseTree.js
```diff
@@ -10,6 +12,9 @@
                root.domUtil);
   }
 })(this, function(ohm, ohmEditor, CheckedEmitter, document, cmUtil, d3, domUtil) {
+  // Fake getComputedStyle() for node (required for unit tests)
+  var getComputedStyle = getComputedStyle || function(elt) { return {}; };
```
pdubroy: This is the purpose of the `initModule` stuff above. You should add a new argument named `getComputedStyle` after `domUtil`, and then pass in the stub version from the unit test itself.

visualizer/index.html
```diff
@@ -55,6 +55,7 @@
       visibility: hidden;
       white-space: nowrap;
       width: auto;
+      font-size: 100%;
```
pdubroy: Nit: The properties are in alphabetical order here...can you stick font-size in the right place?

visualizer/parseTree.js
```diff
@@ -88,11 +92,17 @@
     var measuringDiv = $('#measuringDiv');
     var span = measuringDiv.appendChild(domUtil.createElement('span.input'));
     span.innerHTML = inputEl.textContent;
+
+    measuringDiv.classList.add('expandedInputFont');
```
pdubroy: Could we avoid this by just defining a style for `#measuringDiv .input`?

visualizer/parseTree.js
```diff
@@ -134,7 +144,23 @@
       if (!el._input) {
         continue;
       }
-      el._input.style.minWidth = el.offsetWidth + 'px';
+
+      if (!el.classList.contains('hidden') &&
+          domUtil.closestElementMatching('.collapsed', el) == null) {
+        var elStyle = getComputedStyle(el);
```
pdubroy: Nit: I'd drop the `el` from the beginning of all these names...I don't think it improves the clarity.

visualizer/parseTree.js
```diff
@@ -134,7 +144,23 @@
...
+
+        var elPaddingLeft = parseFloat(elStyle.paddingLeft);
```
pdubroy: I'd just use `parseInt` here, they should all be integer values.

visualizer/style/parseTree.css
```diff
@@ -36,6 +43,10 @@
   background-color: #B5D5FF;
 }
 
+.input {
+  display: inline-block;
```
pdubroy: If you end up creating a `#measuringDiv .input` rule like I suggested above, maybe this can go in there as well? (I'm assuming the measuring is the reason you had to add it here, because otherwise the `#expandedInput span` style should already do this.)

PR #125, opened 2016-07-06
Visualizer: fix of semantics editor

lilyx:
Couple of fixes on semantics editor:
- Fix the argument display issue of `Param` expression
- Fix the semantics editor open issue on Windows/Linux #129


pdubroy:
lgtm

PR #130, opened 2016-07-11
Viz: Change "Force evaluation" submenu to use triangle rather than +/-

pdubroy:
@lilyx, please take a look.

I found the "+" and "-" characters on the "Force evaluation" menu item a bit confusing. I think it makes more sense to use a standard arrow symbol to indicate that there is a submenu.


pdubroy:
Here's what the new version looks like:

screen shot 2016-07-11 at 2 18 13 pm

lilyx:
lgtm

PR #131, opened 2016-07-11
toFailure is missing cases

tonyg:
I noticed this when a program written in a grammar parsed by Ohm had a syntax error involving a `Not` wrapping an `Alt`. When the `Not` was turned into a failure, it delegated to its child pexpr, the `Alt`, which has no `toFailure` implementation.

It can be reproduced by supplying input `"b"` to grammar

    G { start = ~("b" | "c") "d" }

(The supplied test case does exactly this.)

The commit so far only adds a case for `Alt`; really, the proper fix is probably to add all the other missing cases as well.

Or, it could be that this approach is mistaken - it is, after all, rather verbose! The implementation for `Alt` that I have supplied produces error messages of the form `(x or y or z or ...)`, which isn't great. Could an alternative make sense? What would it be, if so?

### DIFF
--- a/src/pexprs-toFailure.js
+++ b/src/pexprs-toFailure.js
@@ -50,3 +50,9 @@ pexprs.Apply.prototype.toFailure = function(grammar) {
 pexprs.UnicodeChar.prototype.toFailure = function(grammar) {
   return new Failure(this, this.toDisplayString(), 'description');
 };
+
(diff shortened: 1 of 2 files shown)


alexwarth:
Thanks, Tony!

We also have to do something similar for `Seq`, but Meixian is already working on it.

PR #135, opened 2016-07-15
Visualizer: Fix default collapse for parameterized applications (e.g. ListOf)

mroeder:
This is a fix for parameterized rules that so far appeared collapsed in the visualizer by default but will now be expanded until a real syntactic rule is hit.
before-after-listof


pdubroy:
Thanks Marko!

PR #137, opened 2016-07-25
Add support for generating examples from a pexpr

sakekasi:


src/pexprs-madlibs.js
```diff
@@ -0,0 +1,209 @@
...
+// --------------------------------------------------------------------
+
+var common = require('./common');
+var pexprs = require('./pexprs');
+
+// --------------------------------------------------------------------
+// Operations
+// --------------------------------------------------------------------
+
+pexprs.PExpr.prototype.generateExample = common.abstract;
```
pdubroy: Shouldn't this be `common.abstract('generateExample')` now?

src/pexprs-madlibs.js
```diff
@@ -0,0 +1,209 @@
...
+    return term.generateExample(grammar, examples, inSyntacticContext, actuals);
+  });
+
+  var categorizedExamples = categorizeExamples(termExamples);
+
+  var examplesNeeded = categorizedExamples.examplesNeeded;
+  var successfulExamples = categorizedExamples.successfulExamples;
+  var needHelp = categorizedExamples.needHelp;
+
+  var returnObj = {};
```
pdubroy: Nit: Our usual name for this kind of variable is `ans`.

test/test-pexprs-madlibs.js
```diff
@@ -0,0 +1,195 @@
...
+    'G {',
+    '  alt = "a" | "b"',
+    '}'
+  ]);
+
+  var random = Math.random;
+
+  Math.random = function() { return 0.6; };
+  var example = g.rules.alt.body.generateExample(g, {}, false, []);
+  t.equal(example.example, 'b');
```
pdubroy: Hmmm, maybe the `example` property should be named `value` instead? It looks a bit weird to write code like `example.example`.

test/test-pexprs-madlibs.js
```diff
@@ -0,0 +1,195 @@
...
+test('Alt', function(t) {
+  var g = makeGrammar([
+    'G {',
+    '  alt = "a" | "b"',
+    '}'
+  ]);
+
+  var random = Math.random;
+
+  Math.random = function() { return 0.6; };
```
pdubroy: Just looking at this test, it's not clear what the significance of 0.6 and 0.2 are. Maybe add a short comment explaining why those values yield `"b"` and `"a"`, respectively?

test/test-pexprs-madlibs.js
```diff
@@ -0,0 +1,195 @@
...
+  var g = makeGrammar([
+    'G {',
+    '  num = "1" .. "9"',
+    '}'
+  ]);
+
+  var example = g.rules.num.body.generateExample(g, {}, false, []);
+  var charCode = example.example.charCodeAt(0);
+
+  t.pass(charCode >= '1'.charCodeAt(0));
```
pdubroy: This is probably not doing what you intended...see https://github.com/substack/tape#tpassmsg. It should probably be `t.ok(...)` or `t.equal(..., true)`.

test/test-pexprs-madlibs.js
```diff
@@ -0,0 +1,195 @@
...
+// test('PExpr');
+//
+test('Alt', function(t) {
+  var g = makeGrammar([
+    'G {',
+    '  alt = "a" | "b"',
+    '}'
+  ]);
+
+  var random = Math.random;
```
pdubroy: You should probably do this in a try/finally block, otherwise if a test throws an exception, you'll never reinstate the real `Math.random`, and it may cause weird errors in other tests.

I.e., make it:

```js
test('whatever', function(t) {
  var random = Math.random;
  try {
    /* test body */
  } finally {
    Math.random = random;
  }
  t.end();
}
```

test/test-pexprs-madlibs.js
```diff
@@ -0,0 +1,195 @@
...
+
+  Math.random = function() { return 0.76; };
+  var example = g.rules.star.body.generateExample(g, {}, false, []);
+  t.equal(example.example, 'aaa');
+
+  Math.random = function() { return 0.1; };
+  example = g.rules.star.body.generateExample(g, {}, false, []);
+  t.equal(example.example, '');
```
pdubroy: This could use a comment to explain why this is expected to produce an empty string here.

src/pexprs-madlibs.js
```diff
@@ -0,0 +1,209 @@
...
+
+pexprs.Lookahead.prototype.generateExample = function(grammar, examples, inSyntacticContext) {
+  return {example: ''};
+};
+
+pexprs.Param.prototype.generateExample = function(grammar, examples, inSyntacticContext, actuals) {
+  return actuals[this.index].generateExample(grammar, examples, inSyntacticContext, actuals);
+};
+
+function repeat(n, fn) {
```
pdubroy: It would be clearer to get rid of this function and just use a `for` loop.

src/pexprs-madlibs.js
```diff
@@ -0,0 +1,209 @@
...
+  if (n < 0) { return; }
+  while (n > 0) {
+    fn(n);
+    n--;
+  }
+}
+
+var generateNExamples = function(that, grammar, examples, inSyntacticContext, actuals, numTimes) {
```
pdubroy: I would rename `that` to `pexpr` here -- it's much more descriptive.

src/pexprs-madlibs.js
```diff
@@ -0,0 +1,209 @@
...
+
+  var categorizedExamples = categorizeExamples(items);
+
+  var examplesNeeded = categorizedExamples.examplesNeeded;
+  var successfulExamples = categorizedExamples.successfulExamples;
+
+  var returnObj = {};
+
+  // it's always either one or the other
+  returnObj.example = successfulExamples.join(inSyntacticContext ? ' ' : '');
```
pdubroy: Hmmm, this should probably be calling `spaces.generateExample` here, since it's possible to override `space` such that `' '` is not a valid example! It's fine to leave that for a follow-up commit though, so maybe just put a TODO in here.

src/pexprs-madlibs.js
```diff
@@ -0,0 +1,209 @@
...
+pexprs.Range.prototype.generateExample = function(grammar, examples, inSyntacticContext) {
+  var rangeSize = this.to.charCodeAt(0) - this.from.charCodeAt(0);
+  return {example: String.fromCharCode(
+    this.from.charCodeAt(0) + Math.floor(rangeSize * Math.random())
+  )};
+};
+
+// these all fail for now
+pexprs.Not.prototype.generateExample = function(grammar, examples, inSyntacticContext) {
+  return {example: ''};
```
pdubroy: Would it make sense to use `null` here instead?

src/pexprs-madlibs.js
```diff
@@ -0,0 +1,209 @@
...
+
+  var successfulExamples = examples.filter(function(example) {
+                                      return example.hasOwnProperty('example');
+                                    })
+                                   .map(function(item) { return item.example; });
+  // could check examples here, here's an example that's invalid
+
+  var needHelp = examples.some(function(item) { return item.needHelp; });
+
+  return {
```
pdubroy: It would be nice to have a comment here explaining what these fields mean. After reading the code, I understand `examplesNeeded`, but don't yet understand `needHelp`.

src/pexprs-madlibs.js
```diff
@@ -0,0 +1,209 @@
...
+pexprs.any.generateExample = function(grammar, examples, inSyntacticContext, actuals) {
+  return {example: String.fromCharCode(Math.floor(Math.random() * 255))};
+};
+
+function categorizeExamples(examples) {
+  var examplesNeeded = examples.filter(function(example) {
+                                  return example.hasOwnProperty('examplesNeeded');
+                                })
+                               .map(function(example) { return example.examplesNeeded; });
+  examplesNeeded = [].concat.apply([], examplesNeeded)
```
pdubroy: I don't get what's going on here. It looks like `examplesNeeded` is always guaranteed to be an Array, so I'm not sure why you need to call `concat`, and anyways `[].concat.apply([], examplesNeeded)` does nothing but copy the contents of `examplesNeeded` into a new array. Am I missing something?

src/pexprs-madlibs.js
```diff
@@ -0,0 +1,209 @@
...
+                            .filter(function(value, index, arr) {
+                              return arr.indexOf(value) === index;
```
pdubroy: I also don't get this...is it supposed to be removing duplicates?

Assuming I understand this code correctly, I think it might be clearer to rewrite it all the map/filter stuff as a single `for` loop. If you need to remove duplicates, you can use an object for that (e.g., `var uniqueExamples = {};` -- then later use `Object.keys` to get an array of the de-duplicated examples).

src/pexprs-madlibs.js
```diff
@@ -0,0 +1,209 @@
...
+
+pexprs.PExpr.prototype.generateExample = common.abstract;
+
+pexprs.any.generateExample = function(grammar, examples, inSyntacticContext, actuals) {
+  return {example: String.fromCharCode(Math.floor(Math.random() * 255))};
+};
+
+function categorizeExamples(examples) {
+  var examplesNeeded = examples.filter(function(example) {
+                                  return example.hasOwnProperty('examplesNeeded');
```
pdubroy: Nit: We don't usually indent like this. To be consistent with other code, it should probably be:

```js
var examplesNeeded = examples.filter(function(ex) {
  return ex.hasOwnProperty('examplesNeeded');
}).map(function(ex) {
  return ex.examplesNeeded;
});
```

pdubroy:
Ok, I left a bunch of comments. Overall this looks pretty good!

One more thing: I'd rename the file to `pexprs-generateExamples.js` -- it's more descriptive, and it's consistent with the name of the method that it defines.

test/test-pexprs-madlibs.js
sakekasi: ah OK. turns out I was reading an outdated article on tape

src/pexprs-madlibs.js
sakekasi: Not and Lookahead generate the empty string as a reasonable default to attempt to generate a valid example (basically, they hope that whatever comes after works according to the encoded constraints). It would be worth it to make this smarter for future work

src/pexprs-madlibs.js
sakekasi: [].concat.apply([], list of lists) flattens the list of lists into a single list. Would it be more clear if I made this into an explicit 'flatten' function?

src/pexprs-generateExample.js
```diff
@@ -8,31 +8,50 @@ var common = require('./common');
 var pexprs = require('./pexprs');
 
 // --------------------------------------------------------------------
+// Helpers
+// --------------------------------------------------------------------
+
+function flatten(listOfLists) {
```
pdubroy: This is better, thanks! I forgot that concat flattened its arguments.

src/pexprs-generateExample.js
```diff
@@ -8,31 +8,50 @@ var common = require('./common');
...
+  return [].concat.apply([], listOfLists);
```
pdubroy: Nit: You can do `Array.prototype.concat.apply` to avoid allocating an extra array that just turns into garbage.

test/test-pexprs-generateExample.js
```diff
@@ -0,0 +1,212 @@
...
+  try {
+    var g = makeGrammar([
+      'G {',
+      '  star = "a"*',
+      '}'
+    ]);
+
+    // Here, a stubbed 'Math.random()' makes 'generateExample()' repeat 'a'
+    //   a certain number of times. for example:
+    //   floor(4 * 0.76) = 3, and floor(4 * 0.1) = 0
```
pdubroy: I'm still not sure what the relevance of '4' is here.

PR #139, opened 2016-07-27
Add example generation web worker to ohm editor

sakekasi:
Adds web worker and worker manager for integration with
editor event system


src/exampleWorker.js
```diff
@@ -0,0 +1,229 @@
...
+// Web Worker that generates examples.
+// communicates with 'exampleGenerationUI.js'
+
+(function(root, initModule) {
+  if (typeof exports === 'object') {
+    module.exports = initModule;
+  } else {
+    initModule(root.ohm, root.utils);
+  }
+})(this, function(ohm, utils) {
```
pdubroy: I think you should change everything above this comment to the following:

```js
/* eslint-env browser */

'use strict';

// Web Worker that generates examples.
// communicates with 'exampleGenerationUI.js'

(function(root, initModule) {
  if (typeof exports === 'object') {
    module.exports = initModule;
  } else {
    initModule(root.ohm, root.utils, self);
  }
})(this, function(ohm, utils, workerGlobalScope) {
  var self = workerGlobalScope;

  self.importScripts('../dist/ohm.js', 'utils.js');
```

Basically, just move the `importScripts` call inside the main function body, and add an extra arg to the main function that allows the `self` to be stubbed out.

That change would make this a valid Node module (right now it throws a ReferenceError), and is a good start at making it testable.

src/exampleWorker.js
```diff
@@ -0,0 +1,229 @@
...
+
+  var overrides = {
+    ident: function(g, ex, syn, args) {
+      return {needHelp: true};
+    },
+    ctor: function(g, ex, syn, args) {
+      return {needHelp: true};
+    },
+    letter: function(g, ex, syn, args) {
+      if (Math.floor(Math.random() * 2) === 0) {
```
pdubroy: Since you use `Math.floor(Math.random() ...)` so often, maybe it's worth making a `randInt(max)` or `randInt(min, max)` function? It would make the code slightly easier to read.

src/exampleWorker.js
```diff
@@ -0,0 +1,229 @@
...
+      }
+    }
+  };
+
+  self.addEventListener('message', function(e) {
+    postMessage(e.data.name);
+    switch (e.data.name) {
+      case 'initialize':
+        /* eslint-disable no-eval */
+        grammar = ohm.makeRecipe(eval(e.data.recipe));
```
pdubroy: I think you should use `JSON.parse(e.data.recipe)` here instead of `eval`.

src/exampleWorker.js
```diff
@@ -0,0 +1,229 @@
+/* eslint-env browser */
+
+'use strict';
+
+self.importScripts('../dist/ohm.js', 'utils.js');
```
pdubroy: Where is utils.js coming from?

src/exampleWorker.js
```diff
@@ -0,0 +1,229 @@
...
+        self.postMessage({name: 'received:neededExamples',
+                          args: [neededExamples]});
+        break;
+      case 'add:userExample':
+        ruleName = e.data.args[0];
+        var example = e.data.args[1];
+        generator.processExampleFromUser(example, ruleName);
+        break;
+    }
+  }, false);
```
pdubroy: Do you need the `false` here? Usually that's the default and can be omitted.

src/exampleWorker.js
```diff
@@ -0,0 +1,229 @@
...
+
+  function addPiecesToDict(trace, examples) {
+    if (trace.expr.constructor.name === 'Terminal') {
```
pdubroy: It's safer to do `trace.expr instanceof ohm.pexprs.Terminal` here. That way, if we ever rename or remove Terminal (unlikely, but it's happened before), it will throw an error.

src/exampleWorker.js
```diff
@@ -0,0 +1,229 @@
...
+      return;
+    } else {
+      if (trace.expr.constructor.name === 'Apply') {
```
pdubroy: Just make this `else if (...) {` rather than `else { if (...) {`.

Also, you should use `instanceof` here too.

src/exampleWorker.js
```diff
@@ -0,0 +1,229 @@
...
+
+  ExampleGenerator.prototype.processExampleFromUser = function(example, optRuleName) {
+    if (optRuleName) {
+      var trace = grammar.trace(example, optRuleName);
+      if (trace.succeeded) {
+        addPiecesToDict(trace, examplePieces);
+      }
+    }
+
+    utils.objectForEach(grammar.rules, function(ruleName) {
```
pdubroy: I don't think you need a helper here...wouldn't `Object.keys(grammar.rules).forEach` do the trick?

I find it's a fine balance as to whether these kinds of things make the code more or less readable. For common stuff like this, my opinion is that it's better to be explicit...and in this case, it doesn't even save any characters. When I see something like this, it makes me think there's a reason that you're not using `Object.keys().forEach` -- so I have to go look at another file to confirm it's doing what I think it's doing. If you just use the obvious thing, then I don't have to think about it at all.

src/exampleWorker.js
```diff
@@ -0,0 +1,229 @@
...
+      return !examplePieces.hasOwnProperty(ruleName);
+    });
+    if (this.examplesNeeded.length < oldSize) {
+      self.postMessage({name: 'received:neededExamples',
+                        args: [this.examplesNeeded]});
+    }
+  };
+
+  ExampleGenerator.prototype.next = function() {
+    var that = this;
```
pdubroy: We usually use the name `self` rather than `that`. No difference, but it's good to be consistent.

src/exampleWorker.js
```diff
@@ -0,0 +1,229 @@
...
+    var currentRuleName = this.rules[this.currentRuleIndex];
+
+    utils.repeat(2, function() {
```
pdubroy: It would be worth explaining why this needs to be run twice, since it's not at all obvious. Also, I think it's better to just use a for loop.

src/exampleWorker.js
```diff
@@ -0,0 +1,229 @@
...
+                               that.examplesNeeded);
+    });
+
+    this.currentRuleIndex = (this.currentRuleIndex + 1) % this.rules.length;
+  };
+
+  function runComputationStep(generator, n) {
+    n = n || 1;
+
+    utils.repeat(n, function() {
```
pdubroy: Just use a for loop here.

src/exampleWorker.js
```diff
@@ -0,0 +1,229 @@
...
+        if (!rules.includes(needed)) {
+          rules.push(needed);
+        }
+      });
+    }
+
+    return examplesNeeded;
+  }
+
+  function parametrized(ruleName, grammar) {
```
pdubroy: Typo: should be 'parameterized'.

src/exampleWorkerManager.js
```diff
@@ -0,0 +1,121 @@
...
+  exampleWorkerManager.registerEvents({
+    'received:examples': ['ruleName', 'examples'],
+    'received:neededExamples': ['neededExamples']
+  });
+
+  var eventsToEmit = ['received:examples', 'received:neededExamples'];
+
+  var exampleWorker = new Worker('exampleWorker.js');
+  /* eslint-disable no-unused-vars */
+  var examplesNeeded = null;
```
pdubroy: Can't you just delete this?

src/exampleWorkerManager.js
```diff
@@ -0,0 +1,121 @@
...
+    exampleWorker.onmessage = onWorkerMessage;
+    exampleWorker.postMessage({
+      name: 'initialize', recipe: grammar.toRecipe()
+    });
+
+    var examples = ohmEditor.examples.getExamples();
+    Object.keys(examples).forEach(function(id) {
+      var example = examples[id];
+      if (example.startRule) {
+        var match = grammar.match(example.text, example.startRule);
```
pdubroy: Nit: Even though this works, I find in these cases it's slightly nicer to move the declaration outside of the loop, so that it's clear what its actual scope is.

Actually, in this case you can just passing `example.startRule` unconditionally, and it will do the right thing if it's null/undefined.

src/exampleWorkerManager.js
```diff
@@ -0,0 +1,121 @@
...
+
+      if (match.succeeded()) {
+        exampleWorkerManager.addUserExample(example.startRule || grammar.defaultStartRule,
+                                            example.text);
+      }
+    });
+  }
+
+  ohmEditor.addListener('parse:grammar', function(_, g, __) {
+    grammar = g;
```
pdubroy: Do you need to store this? It should be always available in `ohmEditor.grammar`, no?

pdubroy:
Good job, this all looks pretty good! For the most part, it's pretty easy to understand since you've used good variable names and kept things pretty simple. One general suggestion is to try to add just a few more comments to your code. We are all guilty of this at times :-) It doesn't have to be much, but a few well placed comments can help get a hold on what the key parts of the code are doing.

Also, this code won't work as-is, right? (E.g., because utils.js is missing.) I think I didn't quite explain myself properly about this, but in general it's ok to check in code that is disabled (or just not currently called from anywhere), but it's best if that code actually *would* work if you called it. For one thing, it makes it easier to review if I have a way of running the code (e.g., "just set `enabled = true` to test this"). Since you're moving everything over to master right now, it's fine to merge this now and fix it up later...but for future PRs, I think it's best to avoid doing it this way.

pdubroy:
Oh, one more thing -- these files should be moved in the `visualizer/` directory.

src/exampleWorker.js
sakekasi: would that work? I thought that the recipe is arbitrary js code

src/exampleWorker.js
sakekasi: in this case, self refers to the global worker context though. would it be OK to alias self in this case?

src/exampleWorker.js
sakekasi: done

src/exampleWorker.js
sakekasi: done

src/exampleWorker.js
sakekasi: done

src/exampleWorker.js
sakekasi: done

src/exampleWorker.js
sakekasi: done

src/exampleWorker.js
sakekasi: done

src/exampleWorker.js
sakekasi: done

src/exampleWorker.js
pdubroy: Oh, good point. Well, if you use a for loop, then you don't need this anyways :-)

sakekasi: done. turns out that I didn't need the 'that' in the first place. But, it would be good to know what to do in this case

src/exampleWorker.js
sakekasi: done

src/exampleWorker.js
sakekasi: done

src/exampleWorker.js
sakekasi: done

src/exampleWorkerManager.js
sakekasi: done

src/exampleWorkerManager.js
sakekasi: done

src/exampleWorkerManager.js
sakekasi: done

src/exampleWorker.js
pdubroy: Not anymore. Marko changed it a few weeks ago to be JSON...and in fact, I think you can just pass the string to `makeRecipe` and it will parse it for you.

sakekasi: ah ok. fixed

src/exampleWorker.js
sakekasi: So, after a bit of thought, in this case, I think I would prefer to use repeat instead of for, just because it's more declarative. There's something similar to repeat strings in common.js. Would that be ok?

pdubroy: Sure, that's fine.

pdubroy:
Cool, looks good, thanks for writing the tests! Left a few more comments, feel free to merge this now and address them in a follow-up commit though.

PR #140, opened 2016-08-02
Add UI to let system request particular examples

sakekasi:
includes:

* adding events, external functions to 'examples.js'
* augmenting example objects with a start rule
* modifying 'splitters.js' to allow splitting of smaller elements than body
* moving css rules from index.html to their own files


visualizer/examples.js
```diff
@@ -6,11 +6,31 @@
...
+    setExample: setExample,
+    setSelected: setSelected,
+    saveExamples: saveExamples
+  });
+
+  ohmEditor.examples.registerEvents({
+    'add:example': ['id'],
+    'set:example': ['id', 'oldValue', 'newValue'],
+    'set:selected': ['id'],
+    'remove:example': ['id']
```
mroeder: Can you put a comment above each line talking about when the event is thrown?
In the existing system, some events fire before, some after the action is/was processed.

visualizer/exampleGenerationRequests.js
```diff
@@ -0,0 +1,167 @@
...
+  var focusedRuleName = '';
+
+  // TODO: clicking off of textarea triggers refresh, deleting content
+
+  // logic for arrow to open drawer and show needed examples menu
+  var examplesNeededDrawer = domUtil.$('#examplesNeededDrawer');
+  var examplesNeededIndicator = domUtil.$('#examplesNeededDrawer .indicator');
+  var exampleSplitter = domUtil.$('#exampleSplitter');
+  examplesNeededDrawer.addEventListener('click', function() {
+    var ExampleRequestContainer = domUtil.$('#ExampleRequestContainer');
```
pdubroy: Nit: Any reason this id and variable begin with a capital letter? If not, should be lowercase.

visualizer/exampleGenerationRequests.js
```diff
@@ -0,0 +1,167 @@
...
+
+      neededExamples.filter(function(ruleName) {
+        return ruleName !== focusedRuleName;
+      }).forEach(function(ruleName) {
+        var li = domUtil.createElement('li');
+        li.appendChild(makeExampleRequest(ruleName));
+        neededExampleList.appendChild(li);
+      });
+
+      if (neededExamples.length === 0) {
```
pdubroy: Nit: This could be `examplesNeededIndicator.classList.toggle('active', neededExamples.length !== 0);`

visualizer/exampleGenerationRequests.js
```diff
@@ -0,0 +1,167 @@
...
+    return request.domNode;
+  }
+
+  // ExampleRequest is a textarea with the ability to check if its contents are valid
+  //  according to a rule. If not, the background is turned pink.
+  function ExampleRequest(grammar, ruleName) {
+    this.ruleName = ruleName;
+    this.grammar = grammar;
+    this.domNode = domUtil.createElement('textarea');
+    this.domNode.setAttribute('type', 'text');
```
pdubroy: Nit: You can just use `this.domNode.type = 'text'` here (and ditto for the `placeholder`).

visualizer/exampleGenerationRequests.js
```diff
@@ -0,0 +1,167 @@
...
+    request.domNode.addEventListener('focusout', function(e) {
+      focusedElement = null;
+      focusedRuleName = '';
+
+      if (request.domNode.value.trim() === '' &&
+          !neededExamples.includes(ruleName)) {
+        request.domNode.parentNode.removeChild(request.domNode);
+      }
+    });
+    request.addListener('validSubmit', function(event) {
```
pdubroy: Is there a scenario where you have multiple listeners for this event? It doesn't look like it, since it looks like `request` doesn't escape this function. In that case, it might be better to just use a callback that you pass to the `ExampleRequest` function.

visualizer/examples.js
```diff
@@ -26,13 +46,20 @@
...
   function checkExample(id) {
     var example = getExample(id);
+    var text = example.text;
+    var startRule = example.startRule;
+    var el = getListEl(id);
     var succeeded;
     try {
-      succeeded = ohmEditor.grammar.match(example).succeeded();
+      if (startRule) {
+        var matchResult = ohmEditor.grammar.match(text, startRule);
```
pdubroy: You can just pass `startRule` unconditionally here -- passing null/undefined is the same as not passing it.

visualizer/examples.js
```diff
@@ -127,16 +187,20 @@
...
+
     examples.forEach(function(ex) {
-      setExample(addExample(), ex);
+      setExample(addExample(), ex.text, ex.startRule);
     });
 
     // Select the first example.
-    var first = domUtil.$('#exampleList li:first-child');
-    var firstId = first ? first.id : -1;
+    var firstIDDOM = domUtil.$('#exampleList li:first-child');
```
pdubroy: Nit: Maybe 'firstEl' or 'firstExample' would be a slightly better name here? (Your call.)

visualizer/examples.js
```diff
@@ -162,6 +226,12 @@
         setExample(selectedId, cm.getValue());
         saveExamples();
       }
+    },
+    'Alt-S': function(cm) { // save (windows)
```
pdubroy: Shouldn't it be `'Ctrl-S'`? And I would factor this function out so that both shortcuts can use the same implementation.

visualizer/style/examples.css
```diff
@@ -0,0 +1,149 @@
...
+  text-align: left;
+  height: 20px;
+  width: 95%;
+}
+
+#exampleRequests ul textarea.invalid {
+  background-color: #FDE0E0 !important;
+}
+
+#exampleRequests ul textarea::-webkit-input-placeholder { /* Chrome/Opera/Safari */
```
pdubroy: I think you should probably just share the same rules for all of these...i.e.:

```
#exampleRequests ul textarea::-webkit-input-placeholder,
#exampleRequests ul textarea::-moz-placeholder,
#exampleRequests ul textarea:-ms-input-placeholder, {
  ...
}
```

pdubroy:
Mostly looks good, just a few comments. Though when I tried it out, I saw a bunch of errors in the console:

screen shot 2016-08-02 at 11 43 41 am
screen shot 2016-08-02 at 11 43 51 am

Looks like maybe `example.text` is undefined on line 42 in exampleWorkerManager.js?

visualizer/index.html
```diff
@@ -156,70 +163,6 @@
...
-      background-color: #f7f7f7;
-    }
-
-    #exampleList code {
-      font-family: Menlo, Monaco, sans-serif;
-      font-size: 12px;
-    }
-
-    #exampleList code:empty::before {
-      content: '\a0';  /* Insert &nbsp; to keep the correct height. */
```
pdubroy: Did you mean to delete this rule? I think we need it, to ensure that the examples keep the same height even when there is no text in them.

visualizer/examples.js
sakekasi: Ctrl-S is bound to save. is there any way to override that in the browser?

visualizer/examples.js
sakekasi: done

visualizer/exampleGenerationRequests.js
sakekasi: done

visualizer/exampleGenerationRequests.js
sakekasi: done

visualizer/exampleGenerationRequests.js
sakekasi: done

visualizer/examples.js
sakekasi: done

visualizer/examples.js
sakekasi: done

visualizer/style/examples.css
sakekasi: done

visualizer/index.html
sakekasi: done

visualizer/examples.js
pdubroy: Yep, normally you just need to call `e.preventDefault()` on the appropriate event.

However, I believe that CodeMirror's "extraKeys" functionality already handles this, because Cmd-S is save on Mac OS, and the existing Cmd-S stuff that's here doesn't trigger a browser save dialog for me.

sakekasi: OK. made the change, but we may have to fix it later on.

sakekasi:
I'll fix that error in the next commit before the pull

visualizer/examples.js
pdubroy: Not sure what you mean. What will we have to fix?

visualizer/exampleGenerationRequests.js
sakekasi: In this particular use of the 'ExampleRequest' class, I don't imagine a situation where another listener would want to subscribe to this event. But, If we were to reuse this element at some point, then it would make sense to have multiple subscribers that listen for when a valid item is submitted

visualizer/examples.js
sakekasi: It did last time I tried it on my linux box. But, I never added an 'e.preventDefault()' call.

visualizer/examples.js
```diff
@@ -205,11 +208,11 @@
 
   // Save the current contents of all examples to localStorage.
   function saveExamples() {
-    var elems = domUtil.$$('#exampleContainer ul li code');
-    var examples = elems.map(function(elem) {
-      return elem.textContent;
-    });
-    localStorage.setItem('examples', JSON.stringify(examples));
+    localStorage.setItem('examples', JSON.stringify(
+      Object.keys(exampleValues).map(function(key) {
```
pdubroy: Can you just stick the result of the map in a local variable named `examples` like it was before? I think it reads more clearly that way.

sakekasi:
merged manually in commit 7c0328d

PR #141, opened 2016-08-08
Add example generation link overlay

sakekasi:


visualizer/style/exampleGenerationLinks.css
```diff
@@ -0,0 +1,61 @@
+.currentExampleLink.coveredExampleLink {
+  text-decoration: underline;
+  color: #268BD2;
+  cursor: pointer;
+  font-weight: bold;
+}
+
+.coveredExampleLink {
```
pdubroy: Looks like there are some redundant rules here, with the `.currentExampleLink.coveredExampleLink` block above. Or am I wrong?

visualizer/style/exampleGenerationLinks.css
```diff
@@ -0,0 +1,61 @@
...
+  text-decoration: underline;
+  /*color: hsl(0, 0%, 50%);*/
```
pdubroy: It's best not to check in commented-out code. If you have a good reason to, it should have a comment explaining why -- otherwise this stuff tends to hang around forever and nobody remembers why :-)

visualizer/exampleGenerationLinks.js
```diff
@@ -0,0 +1,298 @@
...
+  var isMouseDown;
+  var linksEnabled;
+
+  // TODO: spin off common functionality as helpers
+  function isPlatformMac() {
+    return /Mac/.test(navigator.platform);
+  }
+
+  // modifier+Shift
+  function areLinksEnabled(e) {
```
pdubroy: It's bit misleading to call this `areLinksEnabled` -- from that name, I would expect this to be a pure function, rather than to assign into a shared variable. I'd suggest `updateLinksEnabled` instead.

visualizer/index.html
```diff
@@ -467,6 +470,7 @@
 
   <div id="options">
     <ul>
+      <li><input type="checkbox" name="showExampleGenerator"><label>Show Example Generator</label></li>
```
pdubroy: How about "Generate examples" instead? And it should be sentence case to match the others.

pdubroy:
LGTM with the changes I suggested.

visualizer/exampleGenerationLinks.js
sakekasi: done

visualizer/index.html
sakekasi: Is this for the name or the label? The only thing that's confusing about 'generate examples' is that checking the box doesn't actually generate examples. It just shows the UI

visualizer/style/exampleGenerationLinks.css
sakekasi: done

visualizer/style/exampleGenerationLinks.css
sakekasi: done

visualizer/index.html
pdubroy: I meant for the label. Ok, feel free to leave it as "Show Example Generator" but please change to sentence case to be consistent.

pdubroy:
I got this error when I tested in an incognito window: 
screen shot 2016-08-08 at 7 38 04 pm

Do you see the same error on this branch? If so, you should try to fix it before merging this in.

visualizer/index.html
sakekasi: done

sakekasi:
Latest commit should fix the bug

PR #143, opened 2016-08-08
Fix visualization node width bug

sakekasi:


pdubroy:
I just gave this a try, and it still doesn't seem to have fixed the problem:

screen shot 2016-08-09 at 12 53 19 pm

sakekasi:
The latest commit should line up the left sides, but the right sides don't line up because of a transparent border on nodes with the class 'self'

image

This means that the input will always stick out 2px more than its pexpr. I'm not quite sure how to go about fixing this. I'd appreciate any advice you have as to how to go about doing that

pdubroy:
Thanks for this. Would you mind updating it against master? Right now it's really hard to look at the changes -- I'm seeing a lot of stuff that isn't related to your change:

screen shot 2016-08-22 at 4 19 08 pm

Issue #145, opened 2016-08-15
Visualizer: more appropriate  way to display result that represents a CSTNode


lilyx:
When editing an action, if users try to just return a CST node by calling its name, i.e. `return AddExp;`, a node wrapper, instead of the actual CST node, will get returned.

A node wrapper won't nicely displayed, and it will slow down the editor extremely due to the size of its properties.

So for the cases like this, maybe we should display the result as the actual CST node instead of the wrapper? Or we could return a string that indicates it's a wrapper, e.g. "node wrapper of [CST node name]"?

mroeder:
I thought that already is the case, tried it out and if I call `toString()` on the wrapper object that I get back, I already see `[semantics wrapper for G]`.

lilyx:
Yes, in which case you have to call `AddExp.toString()` to get the `[semantics wrapper for G]` displayed. When the editor try to display a result, it calls `JSON.stringify` to display the actual result, instead of taking use of the `toString()`.

mroeder:
I guess then implementing a `toJSON()` for the wrapper makes sense too. It can be similar to the `toString()`. `toJSON()` is called as part of the `JSON.stringify()` protocol.

pdubroy:
Do we really want to display the wrappers as JSON though? It might make more sense to detect them (e.g. with an `instanceof` check) and just call `toString` on them.

lilyx:
That might be a little hard for detecting an array/object that contains the wrappers, like `[wrapper]`, or `{node: wrapper}`. Unless we want to try to eagerly detect them all?

pdubroy:
@lilyx: Ah, good point. Then @mroeder's suggestion makes sense to me.

PR #146, opened 2016-08-16
Add `toJSON` to node wrapper.

lilyx:
This is trying to address #145


src/Semantics.js
```diff
@@ -43,6 +43,11 @@ Wrapper.prototype.toString = function() {
   return '[semantics wrapper for ' + this._node.grammar.name + ']';
 };
 
+// This is used by ohm editor to display a node wrapper appropriately.
+Wrapper.prototype.toJSON = function() {
+  return '[semantics wrapper for ' + this._node.grammar.name + ']';
```
mroeder: I talked to @pdubroy and clarified some of his concerns with the `toJSON`. This seems fine but can you just call the `toString` method here?

lilyx: Oops, I meant to call `toString`! I'll fix it!

PR #149, opened 2016-08-19
Examples: move add button to bottom

acarabott:
Maybe controversial, hence the separate pull request.

Move the add new example button to _below_ the example list, right above the example editor.
Auto scrolls to the newly added example

### DIFF
--- a/visualizer/examples.js
+++ b/visualizer/examples.js
@@ -219,7 +219,9 @@
   // ----
 
   domUtil.$('#addExampleButton').onclick = function(e) {
-    setSelected(addExample());
+    var newExampleId = addExample();
+    setSelected(newExampleId);
+    domUtil.$('#exampleList').scrollTop = getListEl(newExampleId).offsetTop;
   };
 
   var uiSave = function(cm) {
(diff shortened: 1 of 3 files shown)


pdubroy:
Hey Arthur, I think this is a good idea, but I wonder if it might be better to put it _inside_ the list instead and make it a link ("+ Add example"). I'm thinking something like this:

screen shot 2016-08-22 at 4 19 08 pm

What do you think?

acarabott:
Hey Patrick, for me this is a quick stopover design that was quick to implement and saved me time while working on my grammar.
I actually think the example editor should just have inline editing of all examples, with (like you suggest) an "add example" inside the list at the bottom, which is really just an editable example.
Check out my feedback document (emailed to flex list) for a better description.

pdubroy:
Yep, I totally agree on the inline editing. I had always planned on doing that, and the current UI was a quick hack to just get something working.

We can land this stopgap fix if you'd like, but my preference would be to just go ahead and implement the inline editing.

acarabott:
Your call, I have this in my build and I may be the only person actively using it.

In general I think that having _some improvement_ may be detrimental to having the proper design implemented. On the other hand, depends how long it will take to implement the proper version, as this is a definite improvement (imo).

PR #156, opened 2016-09-13
Add VisitorFamily to extras.

pdubroy:
In order to make it easier for people to try this out, this pull request adds my VisitorFamily implementation to `extras/`. Once we've worked any kinks out, we can think about implementing some of the pexpr methods using this.

@mroeder, can you give this a quick review? I will squash this when I merge to master, but for now want to keep the commits separate so as to not lose some of the stuff that is in the earlier commits. For that reason, it's best to view look at the final diff and not the individual commits.


extras/VisitorFamily.js
```diff
@@ -0,0 +1,170 @@
...
+// --------------------------------------------------------------------
+
+// Helpers
+
+function getProp(name, thing, fn) {
+  return fn(thing[name]);
+}
+
+function mapProp(name, thing, fn) {
+  return thing[name].map(fn);
```
mroeder: I don't exactly remember why, but I had to add a check here to get it running with my (potentially) nullable properties (if an AST can have child nodes as property or the property is null):
```javascript
if (!(thing[name] instanceof Array)) {
  return null;
}
```

mroeder:
Left a comment, otherwise LGTM (I have already used it successfully :-)).

extras/VisitorFamily.js
pdubroy: Ok. I went ahead and merged as-is because I want to think about this case a bit more. First, I would think that mapProp should always return an array, so the implementation should be:

```
if (!Array.isArray(thing[name])) {
  return [];
}
```

Second, I'm slightly concerned about typos in the config map leading to bugs. E.g., what should we do if you specify `'foo[]'` and the object has no property named 'foo'?

When I implemented this `[]` shorthand, I was also thinking that it might make sense to use `?` to indicate optional properties. Applied to this case, what about the following:

- `'foo?[]` to mean a property that may or may not exist
- `'foo[]?'` for a property that must exist on the object, but may either be an Array or null/undefined.

Of course, this really starts to resemble Typescript's type declarations; if we took it further, we could have `'foo?: []'` and `'foo: [] | null'` for the two cases. Not the extra complexity is worth it. That said, I do think it's nice to have some stricter checking for stuff like this, because I find that with these kinds of recursive operations, it can be really tricky to figure out where a null/undefined originated from.

Any thoughts?

Issue #162, opened 2016-09-27
Handling case-insensitive keywords


pdubroy:
As pointed out in #157, some languages (e.g. SQL and FORTRAN) have case-insensitive keywords. That is, in SQL, `sElecT` and `SELECT` are equivalent. Currently, supporting this in Ohm requires a separate rule for each letter -- e.g., `select = s e l e c t`. This is awkward an inefficient.

It would be nice to have a better solution -- either a built-in rule for case-insensitive terminals (e.g. `caseInsensitive<"select">`), or else some kind of escape hatch that would enable custom matching to be implemented as a JavaScript function for cases like this.

joshmarinacci:
Escape hatch please!

alexwarth:
I vote for the `caseInsensitive<>` solution because:

1. it will keep the grammar platform-independent (unlike an escape hatch), and
2. its meaning is obvious (unlike adding a special syntax for case-insensitive terminals).

joshmarinacci:
I know, but I keep hoping. We will find more and more cases that need new syntaxes or an escape hatch.

PR #163, opened 2016-10-04
Add built-in rule for case-insensitive string matching (#162)

pdubroy:
@alexwarth: Please take a look.

Before landing this, I'll probably split out everything into the separate `pexprs-` files, just like everything else, but I thought it was helpful to first implement this way, so we have an idea of what it looks like to implement a primitive PExpr from scratch.


src/InputStream.js
```diff
@@ -25,7 +37,13 @@ InputStream.prototype = {
     return this.source[this.pos++];
   },
 
-  matchString: function(s) {
+  matchString: function(s, optIgnoreCase) {
+    if (optIgnoreCase) {
+      var substring = this.sourceSlice(this.pos, this.pos + s.length);
+      var match = regExpForString(s, IGNORE_CASE).exec(substring);
```
alexwarth: Wouldn't it be faster and easier to compare `substring.toLowerCase()` to a lower-cased version of the `CaseInsensitiveTerminal`'s associated string? Then you wouldn't have to worry about regular expressions. (And you could just store the lower-cased string for the terminal in the `CaseInsensitiveTerminal` pexpr, to avoid lower-casing it every time you match with that pexpr.)

src/pexprs-substituteParams.js
```diff
@@ -24,7 +24,6 @@ pexprs.any.substituteParams =
 pexprs.end.substituteParams =
 pexprs.Terminal.prototype.substituteParams =
 pexprs.Range.prototype.substituteParams =
-pexprs.Terminal.prototype.substituteParams =
```
alexwarth: Good catch!

src/InputStream.js
alexwarth: Oh yes, I didn't think about the number of characters in the string -- yay FTW :) But in that case, the RegExp thing wouldn't have worked anyway, because `substring` was computed from `s.length`, right?

And about the caching thing: there are cases where you *would* be able to cache the lower-cased string in the application pexpr -- I'm thinking of when the argument to `CaseInsensitive` is a string literal pexpr (so you know it's always going to be the same thing). But in general, you're right... it could be a variable (one of the formal arguments of the rule from which `CaseInsensitive` is being called) and in that case we wouldn't be able to cache it.

pdubroy: Right, we could cache it in the Apply pexpr. My point was that we can't cache it in the CaseInsensitiveTerminal PExpr.

Do you think I should implement that? Rather than special case CaseInsensitiveTerminal (the Apply.eval() logic is pretty complicated already),  I wonder if we should think about a more general solution for this kind of optimization. It's a bit like constant folding.

PR #167, opened 2016-11-01
Fix for #94

njjewers:
The example now loads from ohm-js if ../.. is not a module


pdubroy:
I've been thinking about this...it's surprisingly tricky to figure out how we should handle it, actually. My worry with the way you've done it here (first requiring `'../..'`), is that if that file is used outside this repository, `'../..'` could be *anything*.

How about just requiring `'ohm-js'`, and then creating an `examples/node_modules` directory, which contains a symlink to the root of the repo?

njjewers:
I've made the requested changes. In addition, I've added a basic test case for the ecmascript module

pdubroy:
Merged. Thanks for doing this (and thanks for the test!).

Issue #53, opened 2016-01-20
Show action call stack when a semantic action is missing


pdubroy:
When there is a missing semantic action, we get an error message like: `Error: Missing semantic action for name in ast operation`...but sometimes the error is that an operation/attribute was invoked on a node that it shouldn't have been. To make it easier to debug these kinds of errors, we should somehow indicate whether the operation/attribute was called from.

PR #168, opened 2016-11-01
Fix for #53

owch:


src/Semantics.js
```diff
@@ -369,8 +370,9 @@ function newDefaultAction(type, name, doIt) {
     } else {
       // Otherwise, we throw an exception to let the programmer know that we don't know what
       // to do with this node.
-      throw new Error(
-          'Missing semantic action for ' + this.ctorName + ' in ' + name + ' ' + type);
+
+      throw errors.semanticErrors(
+          this.ctorName, name, type) ;
```
pdubroy: I think it would be simpler to just use a regular return value here, rather than throwing an exception. Exceptions are appropriate for doing a kind of "non-local return", but in this case I think it would be simpler to just return an error value to the caller. Also, try/catch is a known optimization killer in JavaScript, so we should avoiding putting it in hot paths like this.

However, we need to return a value that can't be mistaken for something that could be returned by the programmer's semantic actions. Here's what I'm thinking:

- in `createError`, we could add a field to every error object, that lets us recognize that it's an error created by Ohm. E.g.:

```
var OHM_ERROR = {};

function createError() {
  ...
  e.type = OHM_ERROR;
  return e;
}
```

And then, down below where we do `thisThing.execute`, we could check if `ans` is an Ohm error:

```
if (ans.type === errors.OHM_ERROR) {
  ...
}
```

Make sense?

src/errors.js
```diff
@@ -177,6 +177,15 @@ function multipleErrors(errors) {
       errors[0].interval);
 }
 
+// ----------------- semantic -----------------
+
+function semanticErrors(ctorName, name, type) {
```
pdubroy: I think `missingSemanticAction` would be a better name here.

pdubroy:
Thanks for making these changes. The one thing left to do is to change from using try/catch to just returning the error as a return value from the semantic action.

src/Semantics.js
owch: If we don't throw the error and just return the error, then how should the error message/stack be printed/handled?

pdubroy: Yes, good question. I was think we could just throw the error only for the top-level wrapper, but after looking at it, it's trickier than I thought.

Let me look at this a bit more and get back to you.

owch: The problem we are facing is that we can't think of a way to detect if we are at the top-level.

pdubroy:
This has turned out to by much more complicated than I thought. However, I ran some benchmarks of your solution, and it doesn't actually seem to affect performance at all (tested using `bin/es5bench`), so I will go ahead and merge this.

Thanks, and sorry that it's taken so long to merge!

PR #170, opened 2016-11-12
adding missing 'require' statement to example in readme, plus link to ohm IDE

sfinnie:
Fir example is missing a `require` statement for ohm-js; generates error as is.  Also added link to the ohm IDE as a great way to get started.


pdubroy:
Merged. Thanks!

PR #171, opened 2016-11-13
substitute arguments for apply when making trace entry

### LINKED ISSUE #52: Visualizer: Parse Tree Visualization does not (always) resolve arguments
For a grammar like:
`G {
  Start = WithArg<"123">
  WithArg<Arg> = Arg ForwardArg<Arg>
  ForwardArg<Arg> = "=" Arg
}`
the visualizer shows the following result when matching `123=123`:
visualizing args - small
The `WithArg` bar shows the supplied argument whereas the `ForwardArg` does not.

In a larger context, the argument sometimes is resolved and sometimes shows up as `$0`, etc.:
visualizing args - large

acslk:
Fixes #52 

This PR resolve the issue that the visualizer (and trace) do not always substitute argument with its supplied value. 

current

In state.eval(expr), expr.eval(state) is called to obtain the result, and then expr is used make the trace entry. The issue occurs because for the Apply expression, substituteParam, which supply the arguments with actual value, is called in expr.eval(state) to make a new Apply expression that is used for the result, while the expr that the state used for tracing is not substituted.

The PR addresses this problem by adding the substitution for Apply in getTraceEntry in State.

### DIFF
--- a/src/State.js
+++ b/src/State.js
@@ -230,6 +230,12 @@ State.prototype = {
 
   // Returns a new trace entry, with the currently active trace array as its children.
   getTraceEntry: function(pos, expr, succeeded, bindings) {
+    if (expr instanceof pexprs.Apply) {
+      var app = this.currentApplication();
+      var actuals = app ? app.args : [];
+      expr = expr.substituteParams(actuals);
+    }
+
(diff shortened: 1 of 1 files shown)


pdubroy:
Merged. Thanks!

Issue #136, opened 2016-07-19
Allow syntactic rule application inside a lexical rule


pdubroy:
As discussed on [this thread](https://groups.google.com/a/cdglabs.org/forum/#!topic/ohm/Vl3PBFn6g58) in the Google Group:

> Take JS's template strings, for example:
> 
> ```
> `The answer is ${6 * 7}`
> ```
> 
> To parse this, it would be nice if you could write something like:
> 
> ```
> tstr
>   = "`" tstrPart* "`"
> 
> tstrPart
>   = "${" Expr "}"      -- expr
>   | ~("`" | "${") any  -- char
> ```
> 
> Right now you can't do this, because that invocation of Expr is not allowed. But it could be.
> 
> Yesterday Pat and I talked about this, and we think that it might be a good idea to allow syntactic rules to be invoked from inside lexical rules -- this would probably require an operator, something like the analog of the "lexification" operator, `#`. For now, I'll use `@`.
> 
> We could rewrite the tstrPart rule above using @ as follows:
> 
> ```
> tstrPart
>   = "${" @Expr "}"      -- expr
>   | ~("`" | "${") any   -- char
> ```
> 
> The semantics of `@` would be the same as the semantics of the application of a start rule that is syntactic. In other words, `@Expr` would skip both leading and trailing whitespace. (That's exactly what you'd want in this example.)

codeZeilen:
Why would you need a special operator for this? Is there any other semantic that an application of a syntactic rule might have at that point?

(Just stumbled upon this as we noticed an inconsistency in Ohm/S which does not restrict users to use lexical and syntactic rules in the bodies of rules of the other kind.)

pdubroy:
@codeZeilen Just seeing this comment now, since this issue came up on the Ohm Discord server. I can't actually remember why we wanted to introduce special syntax for this.

@alexwarth, do you remember? Do you see any problem with allowing syntactic rules to be appear in the body of a lexical rule, without any explicit syntax?

4silvertooth:
> issue came up on the Ohm Discord server.

Here is the grammar

```
TapeGrammar {

Tape 
  = ( #lineComment | Assign | ResultBlock | #op )+
  | end
  
ResultBlock
  = TapeBlock #(result)?

TapeBlock
  = TapeBlock #addop #atom   -- add
  | TapeBlock  #subop  #atom   -- subtract	
  | TapeBlock  #mulop  #atom   -- multiply
[...]

PR #172, opened 2016-11-13
Fixed the lexical and syntactical issue #136.

quackerd:
- Added “@“ operator.


src/pexprs-allowsSkippingPrecedingSpace.js
```diff
@@ -22,6 +22,7 @@ pexprs.PExpr.prototype.allowsSkippingPrecedingSpace = common.abstract(
   Generally, these are all first-order expressions and (with the exception of Apply)
   directly read from the input stream.
 */
+pexprs.Syn.prototype.allowsSkippingPrecedingSpace =
```
pdubroy: I would think that this should be down below, under "higher-order expressions that don't directly consume input" (where you put `Lex` as well). But, TBH I'm not why Lex wasn't already there.

src/pexprs-assertAllApplicationsAreValid.js
```diff
@@ -13,9 +13,11 @@ var pexprs = require('./pexprs');
 // --------------------------------------------------------------------
 
 var lexifyCount;
+var syntactifyCount;
```
pdubroy: Now that we have both `#` and `@`, I think what we need to do here is maintain a stack, just like `inLexifiedContextStack` in the State object. `foo = @(#Bar)` should still result in an error -- it is invoking a syntactic rule from a lexical context.

test/test-ohm-syntax.js
```diff
@@ -1163,6 +1163,41 @@ test('lexical vs. syntactic rules', function(t) {
         /Cannot apply syntactic rule R from here \(inside a lexical context\)/);
   });
 
+    // Unit test for @ operator
+    it('syntactify operator works as expected', function() {
+      var g = makeGrammar(['G {',
+      'tstr = "`" tstrPart* "`"  ',
```
pdubroy: These lines should be indented.

src/pexprs-assertAllApplicationsAreValid.js
```diff
@@ -12,49 +12,62 @@ var pexprs = require('./pexprs');
...
-  lexifyCount++;
+  inLexifiedStackContext.push(true);
   this.expr._assertAllApplicationsAreValid(ruleName, grammar);
-  lexifyCount--;
+  inLexifiedStackContext.pop();
 };
 
 pexprs.Alt.prototype._assertAllApplicationsAreValid = function(ruleName, grammar) {
   for (var idx = 0; idx < this.terms.length; idx++) {
+    inLexifiedStackContext.push(inLexifiedStackContext[inLexifiedStackContext.length-1]);
```
pdubroy: I don't understand why you need to add this here. Only Lex or Syn can change the context, so shouldn't they be the only ones pushing and popping?

src/pexprs-assertAllApplicationsAreValid.js
```diff
@@ -64,8 +77,10 @@ pexprs.Apply.prototype._assertAllApplicationsAreValid = function(ruleName, gramm
     throw errors.undeclaredRule(this.ruleName, grammar.name, this.source);
   }
 
+  var isMeLexifiedContext = inLexifiedStackContext[inLexifiedStackContext.length - 1];
+  var isParLexifiedContext = inLexifiedStackContext[inLexifiedStackContext.length - 2];
   // ...and that this application is allowed
-  if (common.isSyntactic(this.ruleName) && (!common.isSyntactic(ruleName) || lexifyCount > 0)) {
+  if ((common.isSyntactic(this.ruleName) && isMeLexifiedContext) && (!common.isSyntactic(ruleName) || isParLexifiedContext)) {
```
pdubroy: I don't get why you need to look at the 2nd value on the stack. Also, since you now initialize the stack with a correct initial value, you shouldn't need to check `common.isSyntactic(ruleName)`. I'd think it should just be this:

```
if (common.isSyntactic(this.ruleName) &&
    inLexifiedContextStack[inLexifiedContextStack.length - 1]) {
  // throw error
}
```

pdubroy:
This has been addressed by #334

Issue #21, opened 2015-06-03
Improve error messages for invalid escape sequences


pdubroy:
Rather than a generic parse error, we should have a message like "Invalid escape sequence: \p"

PR #173, opened 2016-11-13
Experimental support for #21

njjewers:
A quick implementation of #21. This is accomplished by extending GrammarSyntaxError to provide more relevant messages than the default "expected" messages that Ohm gives. This is, in my opinion, the most appropriate place to put it, as this exception is only used for failure to create user grammars, and this keeps all ohm-specific parse failure messages separate from the rest of Ohm's logic.

I won't lie, I'm not entirely satisfied with it quite yet, especially the changes I made to the visualizer to support taking this from the exception. Perhaps it might be better to integrate this elsewhere.


src/errors.js
```diff
@@ -34,11 +35,40 @@ function intervalSourcesDontMatch() {
...
+  var message;
+  var shortMessage;
+
+  var source = matchFailure.state.inputStream.source;
+
+  // Iterate through the failures, in an attempt to produce a more relevant error message
+  var reason = null;
+
+  var failures = matchFailure.getRightmostFailures();
+  for(var i = 0; i < failures.length; i++) {
```
pdubroy: Nit: Missing space before `(`.

src/errors.js
```diff
@@ -34,11 +35,40 @@ function intervalSourcesDontMatch() {
...
+    var failure = failures[i];
+    if (failure.getPExpr().ruleName === 'escapeChar'){
+      // Failure to match an escape sequence, therefore the grammar contains an invalid escape sequence.
+      var escapeSequence = source.substr(matchFailure.getRightmostFailurePosition(),2);
```
pdubroy: Nit: Missing space after `,`.

src/errors.js
```diff
@@ -34,11 +35,40 @@ function intervalSourcesDontMatch() {
...
+      var reason = "Invalid escape sequence \"" + escapeSequence + "\"";
```
pdubroy: You shouldn't be re-declaring `reason` here -- should just be `reason = ...`

src/errors.js
```diff
@@ -34,11 +35,40 @@ function intervalSourcesDontMatch() {
...
+      break;
+    }
+  }
+  
+  if (reason) {
+      // Use the more relevant reason an the error message
```
pdubroy: Nit: 'as', not 'an'

src/errors.js
pdubroy: Since you call `matchFailure.getRightmostFailurePosition()` several times here, why not declare `var rightmostFailurePos = matchFailure.getRightmostFailurePosition();` above the `for` loop?

Issue #43, opened 2015-09-22
Improve error messages resulting from bad "operation prototype strings", etc.


alexwarth:
Right now if you specify an invalid operation prototype string when you add an operation to a semantics, e.g.,

```
MyGrammar.semantics().addOperation('myOperation(x, y', {...})
```

... you'll get a generic Ohm error message like this:

```
> 1 | myOperation(x, y
                      ^
Expected ")"
```

... which is not very nice.

PR #174, opened 2016-11-13
Issue #43 - Customizable operation parse error

kyuweftea:
@pdubroy @alexwarth Please review


src/util.js
```diff
@@ -146,3 +153,9 @@ exports.getLineAndColumnMessage = function(str, offset /* ...ranges */) {
   }
   return sb.contents();
 };
+
+exports.getLineAndColumnMessage = function(str, offset /* ...ranges */) {
```
pdubroy: I think I'd get rid of the optional `ranges` here. This function can be the simple one that includes the line number and only points to a single offset, and any code that needs more than that can call the other version (which I've suggested you rename to `describeSourceLocation`.

poduncan:
@pdubroy Suggestions were applied. Please review.

test/test-util.js
```diff
@@ -90,49 +91,49 @@ test('getLineAndColumnMessage', function(t) {
 });
 
 test('getLineAndColumnMessage with ranges', function(t) {
```
pdubroy: We should rename this test now, since it's not testing `getLineAndColumnMessage` anymore :-)

test/test-errors.js
```diff
@@ -53,12 +53,28 @@ test('match failure', function(t) {
...
 
+test('syntax error in added operation', function(t) {
+  var g = ohm.grammar('G { start = "a" "b" "c" "d" }');
+
+  try {
+    g.createSemantics().addOperation('myOperation(x, y', {});
+    t.fail('Expected an exception to be thrown');
+  } catch (e) {
+    t.equal(e.message, [
+      'Line 1, col 17:',
```
pdubroy: It doesn't appear that this test passes.

Issue #56, opened 2016-01-31
"Ambiguous" left- and right- recursive rules should be an error


alexwarth:
(See #55 for the backstory.)

Right now it's possible for Ohm programmers to write "ambiguous" left- and right-recursive rules in a grammar, e.g.,

```
AddExp
  = AddExp "+" AddExp  -- plus
  | AddExp "-" AddExp  -- minus
  | MulExp
```

**This should be a compile-time error.**

Why? Because the associativity of the `+` and `-` operators (see above) is not obvious, and it easily could have been. For instance, if you want them to associate to the left, then a much better way to write the above rule is:

```
AddExp
  = AddExp "+" MulExp  -- plus
  | AddExp "-" MulExp  -- minus
  | MulExp
```

And now the associativity is obvious. (Similarly, if right-associativity is desired, then `AddExp` should be rewritten to be right-recursive.)

pdubroy:
I see a few issues with the implementation in #175:

- it only detects direct recursion
- it wouldn't catch the following case:
   ```
   G {
     AddExp = AddExp "+" AddExp noop -- rec
            | digit
     noop =
   }
   ```

A solution for this should probably use the existing mechanism for detecting left recursion, and build on that.

PR #175, opened 2016-11-14
Resolve issue #56. 

akeizer:
Adds new error that fires on compile of left/right associativity ambiguousness. Includes test to check change.

Solution involves finding parent for the appropriate line, passing it in and making a check that verifies the case where there is a left/right associativity issue as per the issue #56. 

I'm part of the UW group.

### DIFF
--- a/src/pexprs-assertProperAssociativity.js
+++ b/src/pexprs-assertProperAssociativity.js
@@ -1,0 +1,60 @@
+'use strict';
+
+// --------------------------------------------------------------------
+// Imports
+// --------------------------------------------------------------------
+
+var common = require('./common');
+var errors = require('./errors');
+var pexprs = require('./pexprs');
+
+// --------------------------------------------------------------------
+// Operations
+// --------------------------------------------------------------------
+
+pexprs.PExpr.prototype.assertProperAssociativity = common.abstract(
+  'assertProperAssociativity'
+);
+
+pexprs.any.assertProperAssociativity =
+pexprs.end.assertProperAssociativity =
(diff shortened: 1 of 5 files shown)


test/test-errors.js
```diff
@@ -251,3 +251,28 @@ test('errors for Not-of-<PExpr>', function(t) {
...
+
+test('Error where left-right associativity is unclear', function(t) {
+  try {
+    ohm.grammar(
+      'G {\n' +
+      '  A = A "+" A -- add \n' +
+      '  | B\n' +
+      '  B = "117"\n' +
+      '}');
+    t.fail('Expected an exception to be thrown');
```
pdubroy: The more idiomatic way to ensure an exception is thrown is to use `t.throws()`: https://github.com/substack/tape#tthrowsfn-expected-msg

src/pexprs-assertProperAssociativity.js
```diff
@@ -0,0 +1,60 @@
...
+pexprs.end.assertProperAssociativity =
+pexprs.Terminal.prototype.assertProperAssociativity =
+pexprs.Range.prototype.assertProperAssociativity =
+pexprs.Param.prototype.assertProperAssociativity =
+pexprs.Lex.prototype.assertProperAssociativity =
+pexprs.UnicodeChar.prototype.assertProperAssociativity = function(ruleName) {
+  // no-op
+};
+
+pexprs.Alt.prototype.assertProperAssociativity = function(ruleName) {
```
pdubroy: You might as well just group *all* of the no-op versions together.

PR #177, opened 2016-12-06
Spec compliant parsing of all possible ES5 identifiers

mroeder:
The grammar for ES5 is missing some rather rare characters that can be used as identifier names, like 䩶 (U+4A76).

It still is different from ES6 which uses the Unicode categories `ID_Start` and `ID_Continue` for `identifierStart` and `identifierPart` (those may be worth including in the `ProtoBuiltInRules`).

Those additional rules add some (more) complexity to the ES5 grammar and also raise the time for parsing ES5 a bit (`es5bench`).


pdubroy:
Can you point me to the relevant part of the ES5 spec?

mroeder:
I guess the relevant part to point you to is https://es5.github.io/#x7.6 and its reference of _UnicodeLetter_. Our `letter` rule is somewhat outdated (I guess) and doesn't seem to include all the characters currently recognized under _“Uppercase letter (Lu)”, “Lowercase letter (Ll)”, “Titlecase letter (Lt)”, “Modifier letter (Lm)”, “Other letter (Lo)”, or “Letter number (Nl)”_ (the latter not being being included in `letter` but in the ES5 grammar itself as `unicodeCategoryNl`).

We cannot parse the character mentioned above as an identifier whereas for example acorn can parse it (with the suggested changes it can).

pdubroy:
Great, thanks for the details.

I think the best way to fix this is to update the built-in `letter` rule. I just tried in a bunch of languages, and the 䩶 (U+4A76) character seems to be a valid identifier in at least the following languages: Java, Rust, Ruby, Python3, C#, Go. That makes me think it makes sense for it to be in the default definition of letter.

When I last looked at this, I don't remember if there was a reason I left `unicodeCategoryNl` out of the built-in `letter` rule. I just checked, and many languages (at least Java, C#, Python3, and Ruby) seem to support identifiers that begin with the character ᛮ (U+16EE).

So, my feeling is:
- "Other letter" (Lo) belongs in the built-in `letter` rule
- "Letter number" (Nl) I'm not sure about, I could go either way but I'd lean slightly towards leaving it as-is for now
- It might be a good idea to add built-in rules corresponding to Unicode `ID_Start` and `ID_Continue`, as several language grammars seem to use that (e.g. Python3, Rust). Might need slightly more investigation before we implement that though.

mroeder:
Sounds like a better way of doing it, @pdubroy. I will close this request and branch and submit a new one with updated Unicode rules. Thanks!

PR #178, opened 2016-12-08
Update Unicode categories for Ohm rules

mroeder:
Let's try this again: Here is an update for the `third_party/UnicodeCategories.js` file that pulls in updated regular expressions for the unicode categories that we currently use.
Instead of the current definition, we could:
a) Copy the regular expression directly into the `third_party` file (as before), or
b) Get rid of the `third_party/UnicodeCategories.js` file completely and reference the categories directly.

The current way of importing only includes the necessary files in our browserify package and minified version of it, so that the size of the `dist` files barely changes.

With the current categories, we still would need additional custom character matching in the ES6 grammar (and others) to be able to match all valid identifiers based on `ID_Start ID_Continue*`.

Thoughts @pdubroy, @alexwarth?


pdubroy:
Seems like a good idea to let someone else deal with updating the list when necessary ;-)

The way you've done it here won't quite work if people install ohm-js in "user-mode", right? Because it's a devDependency, but UnicodeCategories.js still `require`s it. I'm not sure why, but the `unicode-9.0.0` package does say this:

> Note: unicode-9.0.0 is supposed to be used in build scripts (i.e. as a devDependency), and not at runtime (i.e. as a regular dependency).

If we don't want it as a full dependency, I guess we would need to add a build script that will generate the contents of `third_party/UnicodeCategories.js`, without depending directly on the `unicode-9.0.0` package. Something like `browserify third_party/UnicodeCategories.js --bare -s UnicodeCategories` maybe?

mroeder:
Oops, I didn't mean to put it into dev dependencies but just dependencies.
I am not too concerned about the note on the unicode-9.0.0 project and putting it into dependencies would work for me. Objections?

pdubroy:
I decided to check with Mathias about why he recommends only using that package as a dev dependency: https://github.com/mathiasbynens/node-unicode-data/issues/43. Turns out it's mainly because of the size of the package (141 MB, whereas Ohm's current dependencies are about 3MB).

So, I don't know. On one hand it would be nice to keep the dependencies small, but I'm not sure how much it really matters. If you want to land this as is, I can look into writing the build-time script later this week.

One thing I would like to do though, is lock the dependency to a specific version, i.e. use `"unicode-9.0.0": "0.7.0"` without the `^`. Otherwise I worry that the exact same version of the Ohm repo could support different semantics for `letter`, just depending on the exact version of the Unicode library that got installed. (And now that I think of it, that's another reason to prefer a build-time dependency, since then we'd be depending on those regexps "by value" instead of "by reference" ;-))

mroeder:
Change the original commit do include a copy of the regexps instead.
Still working on `identStart`, `identContinue` and `ident` but it seems a bigger change (both in terms of size and in terms of changes in Ohm -- some of our default grammars, like Ohm itself, already have an `ident` rule). Stay tuned!

pdubroy:
LGTM...should we merge this then, and do the other stuff in a separate commit?

mroeder:
Yes, let's do that.

PR #179, opened 2016-12-12
Mention \x and \u escapes in the syntax reference

rtoal:
Thought it might be nice to mention that these escape forms are supported, since the existing language seems to suggest the ones explicitly listed were the only permissible forms.


pdubroy:
Thanks Ray!

PR #187, opened 2017-02-22
Typescript type declarations

JonathanMEdwards:
Not sure how to test this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
   "bugs": "https://github.com/harc/ohm/issues",
   "main": "src/main.js",
   "bin": "src/ohm-cmd.js",
+  "types": "src/ohm-js.d.ts",
   "browserify": {
     "options": "src/main.js --debug --standalone ohm -o dist/ohm.js"
   },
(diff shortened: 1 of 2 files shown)


JonathanMEdwards:
I'm thinking that what really needs to be tested is that the manifest is correct, so we would need to build the ohm npm package then install it into a TypeScript package. Seems like a lot of work.
A simpler alternative would be to published the typings as a separate package at DefinitelyTyped, which TypeScript knows how to automatically access.

pdubroy:
I added a small test in 1b014d9, and tweaked the file a bit to get things working. Let me know if something doesn't look right. I don't fully understand the differences between all the different ways to declare/export the top-level `ohm` object, but this allows `import ohm from 'ohm-js'` to work, which seems reasonable to me.

pdubroy:
Also I moved the declarations into `index.d.ts` in the root of the repo, as this seemed to be the standard way to do it from what I can tell.

JonathanMEdwards:
As I worried, this doesn't work once packaged in NPM.
Typescript now fails to load ohm at runtime, although it is happy at compile time. I have no clue how to debug this. Module loading is a complete mystery.
Also, none of the interfaces are exported, but that is easy to fix.

JonathanMEdwards:
Solved it. Also need to tighten up test as it isn't checking for missing declarations. Will submit PT shortly

PR #189, opened 2017-02-24
Allow both default and named exports. Export the interfaces too. Fixe…

JonathanMEdwards:
…d/added declarations. Tightened up the test to report type inference errors.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,10 +1,12 @@
+export = ohm;
+export as namespace ohm;
 declare namespace ohm {
 
   /**
    * Instantiate the Grammar defined by source. If specified, namespace is
    * the Namespace to use when resolving external references in the grammar.
    */
-  export function grammar(source: string, namespace?: Namespace): Grammar;
+  function grammar(source: string, namespace?: Namespace): Grammar;
 
   /**
    * Create a Grammar instance from the contents of a <script> tag.
@@ -14,7 +16,7 @@ declare namespace ohm {
    * If specified, namespace is
    * the Namespace to use when resolving external references in the grammar.
    */
-  export function grammarFromScriptElement(node?: Node, namespace?: Namespace): Grammar;
+  function grammarFromScriptElement(node?: Node, namespace?: Namespace): Grammar;
 
   /**
    * Create a new Namespace containing Grammar instances for all of the
@@ -22,7 +24,7 @@ declare namespace ohm {
    * If namespace is specified, it will be the prototype of the new
    * Namespace.
    */
-  export function grammars(source: string, namespace?: Namespace): Namespace;
+  function grammars(source: string, namespace?: Namespace): Namespace;
 
   /**
    * Create a new Namespace containing Grammar instances for all of the
@@ -32,7 +34,7 @@ declare namespace ohm {
    * If namespace is specified, it will be the prototype of the new
    * Namespace.
    */
-  export function grammarFromScriptElements(
+  function grammarFromScriptElements(
     nodeList?: NodeList,
     namespace?: Namespace): Namespace;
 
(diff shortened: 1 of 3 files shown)


pdubroy:
Thanks! BTW is it normal in the Typescript world to have both a default and a named export? Or should we just get rid of the default one and use a named export?

JonathanMEdwards:
Actually it looks like I broke the default export. I can't get both default and named working. And typescript won't load the default style. Modules are a total mess. I ended up copying the style from the react definitions. At least it works.

pdubroy:
Yeah, I don't really care too much, just happy if something works.

pdubroy:
I made some changes so that we now actually _execute_ test-typings.ts (what a concept). From what I could tell, the `export as namespace` wasn't doing anything for us.

I tried the following things:

```
import ohm from '..';
import {ohm} from '..';
```

and neither of them work, with or without that `export as namespace` line. But the following still works:

```
import * as ohm from '..';
```

I also installed `@types/tape`, and importing it seems to only work in the same way:

```
import * as test from 'tape';
```

From what I can tell (e.g. by reading [Typescript React & Webpack doc](https://www.typescriptlang.org/docs/handbook/react-&-webpack.html)), this seems to be the "standard" way to import CommonJS modules from Typescript. Does that sound right to you?

If I'm missing something and that line is useful for something, just let me know and I'll put it back in. Either way, I'm not going to think about this stuff any more -- what a mess!

JonathanMEdwards:
Agreed, namespace export is unnecessary. I just copied it from declaration files I saw. And yes, import * and import {...} does seem to be how it's done in Typescript. I just hope that someday we reach the holy land of es6 modules. Thanks for your help.

bd82:
"export as namespace" is only used to export libraries using [UMD](https://github.com/umdjs/umd).
https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#support-for-umd-module-definitions

But it does seem like the concatenated version of ohm is using something similar to umd.
https://unpkg.com/ohm-js@0.13.1/dist/ohm.js

So this pattern may be needed to support users "importing" using namespaces.
```typescript
// es6 imports --> can be transformed to commonjs/amd/systemjs? imports.
import {x, y, z} from "ohm"

// namespace imports - ohm "namespace" globally available
var x = ohm.x
```

pdubroy:
@JonathanMEdwards, no problem, thanks for the PR. I'm happy to have this and to have had a chance to learn a little bit about Typescript.

@bd82: interesting, thanks. The bundled version of Ohm is built using browserify, but I would think that anyone using Typescript should be using the Ohm NPM module, right? So it shouldn't really matter what the browserified version of Ohm is like. (Correct me if I'm wrong.)

bd82:
> but I would think that anyone using Typescript should be using the Ohm NPM module, right? So it shouldn't really matter what the browserified version of Ohm is like. (Correct me if I'm wrong.)

I suspect you are correct, but not because TypeScript users will be installing Ohm via npm,
but because most **modern** TypeScript code bases will be using the ES6 Modules Import/Export syntax
which would be compatible with your definitions.

The problem may arise with **legacy** (amazing how in the JS world 2 years is already ancient 😄 )
TypeScript projects which still use the old "namespaces".

I guess this is why the export as namespace is [commonly used](https://github.com/DefinitelyTyped/DefinitelyTyped/search?utf8=%E2%9C%93&q=%22export+as+namespace%22&type=Code) on d.ts files.

It also seems like (if I understand the links below correctly) that some other d.ts files **manually** export in **both** module formats.
* grunt declared as a [namespace](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/9dad6b2b06440e2dbe7a498931e7f5e70f858c2b/grunt/index.d.ts#L117).
* grunt exported as [module](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/9dad6b2b06440e2dbe7a498931e7f5e70f858c2b/grunt/index.d.ts#L1326).

bd82:
@pdubroy 
Perhaps this is off-topic (should I open an issue?) 
But did you consider migrating the ohm project itself to TypeScript?
This way your definitions will not have a separate lifecycle but instead be generated directly
from the source code, not to mention you will be able to modernize your codebase with ES6+ features.

pdubroy:
@bd82 Maybe someday. I'd like to learn more about Typescript first, and in Ohm there is a long list of things that I'd rather implement first.

PR #218, opened 2018-04-26
Fixes undefined errors with nested expression toFailure(...)

stanistan:
👋 

While trying to get good error messages (via `MatchResult.message`) in a project I'm working on I noticed a bug in nested `pexpr.toFailure(..)` calls where not all of them would pass the grammar through to their child expressions.

```
  TypeError {
    message: 'Cannot read property \'rules\' of undefined',
  }

  Apply.pexprs.Apply.toFailure (node_modules/ohm-js/src/pexprs-toFailure.js:46:29)
  node_modules/ohm-js/src/pexprs-toFailure.js:59:50
  ...
```

There is test case with a complex enough grammar that would hit something like this (there is probably a simpler version of it?) as a regression test.


stanistan:
Thx!

Issue #220, opened 2018-05-03
Problem with NodeList in index.d.ts in node.js project


Gamadril:
I have a node.js based TypeScript project. ohm is installed using npm. Running `tsc -p` throws an error:
`node_modules/ohm-js/index.d.ts(39,16): error TS2304: Cannot find name 'NodeList'.`

I guess because NodeList is browser specific. Commenting out the `grammarFromScriptElements` function signature in index.d.ts makes it compile again.

pdubroy:
I'd like to improve our TypeScript support, so thanks for filing this. Could you share the compiler config you are using (e.g. `tsconfig.json`)?

Gamadril:
sure. it's part of a VS Code extension, I'm currently working on.
```
{
    "compilerOptions": {
        "module": "commonjs",
        "target": "es6",
        "outDir": "build",
        "lib": [
            "es6"
        ],
        "sourceMap": true,
        "noUnusedLocals": true,
        "strict": true
    },
    "exclude": [
        "node_modules",
        ".vscode-test"
    ]
}
```

pdubroy:
Thanks. If you change your `"lib"` entry to have the value `["es6", "dom"]`, does that fix your problem?

I don't fully understand how TS resolves the types in this situation, but what happens if you add `/// <reference types="dom" />` to the top of Ohm's `index.d.ts` (and leave your `"lib"` entry as it is now)?

Gamadril:
["es6", "dom"] seems to fix it, but it also affects the validation of the own project code.

Adding `/// <reference types="dom" />` leads to another error:
`node_modules/ohm-js/index.d.ts(1,23): error TS2688: Cannot find type definition file for 'dom'.`

![image](https://user-images.githubusercontent.com/3903793/39579831-f93c230e-4ee7-11e8-91bf-f57cd8655e87.png)

pdubroy:
Hmm, ok. So for your use case, you'd prefer that the Ohm definitions don't refer to any types from the DOM library, right?

I think maybe the solution is to do something similar to what [ProtoBuf.js does](https://github.com/dcodeIO/ProtoBuf.js/): you should be able to choose to use a subset of the API that doesn't include any browser-related features. E.g., `import * as ohm from 'ohm-js/no-dom';`.

Gamadril:
That would be great.

Gamadril:
[...]

PR #224, opened 2018-07-04
Updated MatchResult declarations

sanboxrunner:
Previously: Missing getRightmostFailurePosition and getRightmostFailures.

Now: updated to have those two methods.

Additionally, corrected spelling error of `iff` to `if`. 

I used the text from the documents for the JSDOC comments.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -34,9 +34,9 @@ declare namespace ohm {
    * If namespace is specified, it will be the prototype of the new
    * Namespace.
    */
-  function grammarFromScriptElements(
-    nodeList?: NodeList,
-    namespace?: Namespace): Namespace;
+  // function grammarFromScriptElements(
+  //   nodeList?: NodeList,
+  //   namespace?: Namespace): Namespace;
 
   /**
    * Create a new namespace. If props is specified, all of its properties
@@ -135,14 +135,27 @@ declare namespace ohm {
    */
   interface MatchResult {
     /**
-     * True iff match succeeded
+     * True if match succeeded.
      */
     succeeded(): boolean;
 
     /**
-     * True iff match did not succeed
+     * True if match did not succeed.
      */
     failed(): boolean;
+
+    /**
+     * If match failed, returns the index in the input stream at
+     * which the match failed.
+     */
+    getRightmostFailurePosition(): number;
+
+    /**
+     * If the match failed returns an array of Failure objects,
+     * describing the failures that occurred at the rightmost failure
+     * position.
+     */
+    getRightmostFailures(): Array<object>;
 
     /**
      * If match failed contains an error message indicating where and


index.d.ts
```diff
@@ -34,9 +34,9 @@ declare namespace ohm {
    * If namespace is specified, it will be the prototype of the new
    * Namespace.
    */
-  function grammarFromScriptElements(
-    nodeList?: NodeList,
-    namespace?: Namespace): Namespace;
+  // function grammarFromScriptElements(
```
pdubroy: Why is this commented out now?

index.d.ts
```diff
@@ -135,15 +135,28 @@ declare namespace ohm {
    */
   interface MatchResult {
     /**
-     * True iff match succeeded
+     * True if match succeeded.
```
pdubroy: How about "Returns true if the match succeeded, otherwise false."

index.d.ts
```diff
@@ -135,15 +135,28 @@ declare namespace ohm {
...
      */
     succeeded(): boolean;
 
     /**
-     * True iff match did not succeed
+     * True if match did not succeed.
```
pdubroy: How about "Return true if the match failed, otherwise false."

pdubroy:
BTW, if you are using Ohm with TypeScript I'd love to hear if how your experience has been — my email address is in my profile.

I have actually been working on porting all of Ohm to TypeScript. It's a work in progress but if you're interested you can find it in this branch: https://github.com/harc/ohm/tree/ts-conversion.

sanboxrunner:
1. Regarding the commented out area: it references `NodeList` which isn't itself declared. I'm not sure what NodeList was, so I couldn't/didn't add a declaration myself for it. I didn't mean to submit this with it comment out -- we can change that back. It's just been throwing errors at me in my projects so I keep it commented out (since I'm in Nodejs, I don't really need it). 

2. I'll adjust those (or do you adjust them? Sorry this is both the first time I've used Github and the first pull request I've ever done in an open source project!)

2. Since I haven't explored Ohm internals too much, using it in Typescript has frankly been a breeze. I am using it to produce a Language Server.

pdubroy:
> Regarding the commented out area: it references NodeList which isn't itself declared. I'm not sure what NodeList was, so I couldn't/didn't add a declaration myself for it. I didn't mean to submit this with it comment out -- we can change that back. It's just been throwing errors at me in my projects so I keep it commented out (since I'm in Nodejs, I don't really need it).

Ah, right. Someone already raised this issue in #220. We should fix that, but yes, for now please uncomment that section for the purposes of this PR.

> I'll adjust those (or do you adjust them? Sorry this is both the first time I've used Github and the first pull request I've ever done in an open source project!)

Cool! Happy to accept your first PR :-)

The usual way to do it would be that you make the changes in your branch, commit them, and then do a `git push`. The PR will be automatically updated to reflect the changes. Then, what I usually to do is to go through each of the comments from the reviewer and comment "Done" (or explain why the suggestion doesn't work, etc.) This way it's clear that everything has been considered and/or addressed.

> Since I haven't explored Ohm internals too much, using it in Typescript has frankly been a breeze. I am using it to produce a Language Server.

Great, glad to hear that. BTW I will try to address #220 ASAP since it seems like at least a few people are running into the problem.

PR #239, opened 2019-02-28
Dependency updates

lguzzon:


mikeaustin:
I’m just starting to use Ohm, but I’m curious why the removal of space in function blocks, such as function() {return 0;} That doesn’t seem like conventional syntax. Also, the merge is named “Dependency updates” but has formatting. Personally, I believe that very short, very specific PRs make them easier to read, approve, and manage (history, reverting, etc.). Last, it’s good to be up to date with modules, but it’s nice to know what changed?

PR #264, opened 2020-09-23
Convert to more modern JS

### LINKED ISSUE #263: Semantics.toRecipe() produces invalid code when actions use method shorthand
Example:

```
~/dev/harc/ohm/ohm-js$ node
Welcome to Node.js v14.4.0.
Type ".help" for more information.
> const ohm = require('.')
undefined
> const g = ohm.grammar('G { start = }')
undefined
> const s = g.createSemantics().addOperation('test', { end(_) {} })
undefined
> eval(s.toRecipe())
Uncaught SyntaxError: Unexpected token '{'
```

The problem is that `toString` produces a different result when the method shorthand is used (`{ o() { } }`) vs using a function (`{ o: function() {} }`).

pdubroy:
- Switch to TypeScript + Webpack for transpiling and bundling JavaScript sources. This allows us to adopt modern JS syntax while keeping the bundles compatible with older browsers. It also opens the path to gradually migrate to typed `.ts` sources.
- Fix #263: Semantics.toRecipe() breaks when actions use method shorthand
- Use ESLint's `--fix` option to automatically modernize the codebase in the following areas:
    * [x] Replace `var` with `let`/`const`
    * [x] Use object property/method shorthand (including in semantic action dictionaries)
    * [x] Use arrow functions
- To do in the future:
    * [ ] Rest parameters
    * [ ] Destructuring
    * [ ] Classes


pdubroy:
fyi @alexwarth

alexwarth:
Looks great to me, @pdubroy!!

pdubroy:
Thanks @alexwarth!

Issue #265, opened 2020-10-27
Add something like `=+` to match the existing `+=`


tonyg:
Currently, `+=` *prepends* a PEG to the supergrammar's rule.

We have been building a parser that uses inheritance to extend a grammar with error recovery productions, and find ourselves wishing for something like a `=+` that *appends* to the supergrammar's rule:

    SomeNT =+ SomeErrorRecoveryPEG
    // ... would be nice if it expanded to...
    SomeNT = Rules | From | The | SuperGrammar | SomeErrorRecoveryPEG

Would a patch implementing a `=+` operator be welcome?

Building on this question, if it were considered valuable to have both `+=` and `=+`, would it make any sense to have `+=` *append* and `=+` *prepend*?

Finally, if `=+` is a bad choice of operator, perhaps `|=` is a better choice for the appending variant?

pdubroy:
Hi @tonyg!

I can see how this would be useful, and it's definitely something we have considered before. It would be nice if we could support this one way or another. That said, I do want to be cautious about adding new syntax.

In the past, we've also talked about supporting "super-send" — i.e., the ability to use `super` to apply the supergrammar's version of the current rule. E.g., `keyword += class` would be sugar for `keyword := class | super`. If we had that, I think you could then solve your case via `SomeNT := super | SomeErrorRecoveryPEG`. What do you think?

tonyg:
I think having super-send is a great idea, yes! ... and could open the door even to *removing* (or deprecating, I suppose) the `+=` syntax :-)

Perhaps it should be spelled something like `^`, or other definitely-not-a-nonterminal, though? Presumably all this has already been discussed back and forth somewhere.

alexwarth:
Hey @tonyg and @pdubroy!

The issue with "super" in Ohm is that it complicates the story of semantic actions. E.g., suppose we have the following grammars:

```
Parent {
  Foo = "foo"
}
```

```
Child <: Parent {
  Foo := ^Foo "bar"
}
```

I think we can all agree that the arity of `Child`'s `Foo` rule is 2, because we have the`^Foo` and the `"bar"`. So how do we specify how the value of `^Foo` is computed?

* Would the "semantic action dictionary" that we provide to `addOperation` need two entries for `Foo`? (One for `Parent`, and another for `Child`?)
* Or can we write a single semantic action that decides what to do based on `arguments.length`? (Ewwwww.)
[...]

PR #267, opened 2020-11-22
feat: implement "super-splice" operator (`...`) (#265)

pdubroy:
- Implements the "super-splice" operator (`...`), which can appear as a top-level term in the body of an overriding rule. Using `...`, a subgrammar can add new cases _after_ or _around_ the supergrammar's implementation, whereas `+=` only allows adding new cases at the beginning.
- TODO before merging:
    * [x] add test for using super-splice when overriding a non-existent rule
    * [x] full support for recipes
    * [ ] ~(maybe) merge `Extend` and `Splice`~


pdubroy:
ptal @alexwarth @tonyg

ohm-js/test/test-ohm-syntax.js
```diff
@@ -976,6 +976,47 @@ test('inheritance', t => {
...
+  t.equal(g.match('@', 'letter').succeeded(), true);
+  t.equal(g.match('a', 'letter').succeeded(), true);
+
+  g = ohm.grammar('G { letter := ... | "@" }');
+  t.equal(g.match('@', 'letter').succeeded(), true);
+  t.equal(g.match('a', 'letter').succeeded(), true);
+
+  g = ohm.grammar('G { letter := "3" | ... | "@" }');
+  t.equal(g.match('@', 'letter').succeeded(), true);
+  t.equal(g.match('3', 'letter').succeeded(), true);
```
alexwarth, line 990: Nit: we should add a test to make sure that `letter` matches `'a'`, too.

ohm-js/test/test-ohm-syntax.js
```diff
@@ -976,6 +976,47 @@ test('inheritance', t => {
...
+      /Expected "}"/,
+      "it's not allowed when extending");
+
+  t.throws(
+      () => ohm.grammar('G { letter := "@" "#" | ... }'),
+      /inconsistent arity/,
+      'it ensures consistent arity');
+
+  /*
+    TODO:
```
alexwarth, line 1010: I suspect the grammar will break if someone uses more than one `...` in an override body. It would be good to catch this and display a nice error message.

ohm-js/test/test-ohm-syntax.js
```diff
@@ -976,6 +976,47 @@ test('inheritance', t => {
   t.end();
 });
 
+test('override with "..."', t => {
+  let g = ohm.grammar('G { letter := "@" | ... }');
+  t.equal(g.match('@', 'letter').succeeded(), true);
+  t.equal(g.match('a', 'letter').succeeded(), true);
```
alexwarth, line 982: We should add another test that ensures that the "before terms" get evaluated before the "...". Something like this might do the trick:

```
let g = ohm.grammar('G { letter := "" | ... }');
t.equal(g.match('', 'letter').succeeded(), true);
t.equal(g.match('a', 'letter').succeeded(), false);
```

ohm-js/test/test-ohm-syntax.js
```diff
@@ -976,6 +976,47 @@ test('inheritance', t => {
...
+
```
alexwarth, line 983: ... and another test that ensures that the "after terms" get evaluated after the "...". How about this?

```
let g = ohm.grammar('G { letter := ... | "ab" }');
t.equal(g.match('a', 'letter').succeeded(), true);
t.equal(g.match('ab', 'letter').succeeded(), false);
```

alexwarth:
Very nice, @pdubroy! I left a couple of comments, but overall it LGTM!

ohm-js/test/test-ohm-syntax.js
pdubroy: Done.

ohm-js/test/test-ohm-syntax.js
pdubroy: Good catch. I had a check but it must have disappeared in refactoring.

ohm-js/test/test-ohm-syntax.js
pdubroy: Nice! Added.

ohm-js/test/test-ohm-syntax.js
pdubroy: Done.

pdubroy:
Thanks for the review @alexwarth!

pdubroy:
Ok, I cleaned up the final TODOs and I think this is ready to merge now.

PR #275, opened 2020-12-28
Modernize Operators Example

rtoal:
Use ES module import/export syntax (type: module in package.json), use spreads instead of concat, and clean up some parameter names.

### DIFF
--- a/examples/operators/README.md
+++ b/examples/operators/README.md
@@ -38,4 +38,4 @@ This example shows how to write the same grammar all four ways, with an associat
 
 The idea is to give grammar designers a look into the various alternatives to pick a style that’s right for them. Even though the first of the four will generally be preferred.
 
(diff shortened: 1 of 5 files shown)


pdubroy (approved):
Woo, thanks Ray!

pdubroy:
Hmmm, tests are failing though, due to the use of `import` and `export`: https://travis-ci.org/github/harc/ohm/builds/751856805#L228. I wonder why `"type": "module"` didn't do the trick?

pdubroy:
Thanks Ray, I merged this in manually along with a fix for the CI issue. Man, using ES modules is still super complicated. The problem here was that ESLint doesn't seem to look at the package.json — it needs to be configured separately to treat modules as ES modules.

rtoal:
Sorry it does not work with Travis. It works on the command line when you do npm test.

PR #284, opened 2021-01-13
feat: add `getLineAndColumn()` method to `Interval`

stagas:
This PR resolves https://github.com/harc/ohm/issues/281.

Notes: the test provided only asserts successful integration with `util.getLineAndColumn()` and the returning object shape -- not meant to be exhaustive as this should be done in `test-util` (currently it is only being tested indirectly by `util.getLineAndColumnMessage()` exhaustive unit tests).

### DIFF
--- a/ohm-js/src/Interval.js
+++ b/ohm-js/src/Interval.js
@@ -49,6 +49,10 @@ Interval.prototype = {
     return new Interval(this.sourceString, this.endIdx, this.endIdx);
   },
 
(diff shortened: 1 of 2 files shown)


pdubroy (approved):

pdubroy:
I'll go ahead and merge this since it's strictly an improvement. It would be great if you can implement the `toString()` in a follow-up PR though!

Issue #281, opened 2021-01-12
Expose getLineAndColumn?


stagas:
We are exposing [`util.getLineAndColumnMessage()`](https://github.com/harc/ohm/blob/f0a893440a151aa04952ac15bd8d7f4b4b824aca/ohm-js/src/util.js#L112) in the `Interval` instances, but I think it would be useful to expose [`util.getLineAndColumn()`](https://github.com/harc/ohm/blob/f0a893440a151aa04952ac15bd8d7f4b4b824aca/ohm-js/src/util.js#L56) as well. I can prepare a PR if necessary. Thoughts?

https://github.com/harc/ohm/blob/f0a893440a151aa04952ac15bd8d7f4b4b824aca/ohm-js/src/Interval.js#L52-L55

pdubroy:
@stagas Sounds reasonable to me. Where are you thinking of putting it? My first thought would be to put it in `extras`.

I assume you are thinking that this is useful for language authors (?)

stagas:
I was thinking just adding a method right next to that one in the `Interval`, exactly the same except it calls `util.getLineAndColumn()`. It's useful when doing your own reflections, managing the way you display the error in an editor or elsewhere but there could be more cases and since we're already doing it for `util.getLineAndColumnMessage()`. Okay, now that I look up the docs I see that `Interval` isn't documented - which is what I was accessing inside the visitor from `.source`, so maybe these are not meant to be public, but I couldn't find any other way to get to the line and number properties but only by adding that method. Is there another way to go about it maybe?

pdubroy:
Sorry, I misunderstood. Sounds good to me, and a PR would be very much appreciated!

IMO `Interval` can be considered public. I don't think it was a deliberate choice not to document it, just that our documentation is not exhaustive :-)

stagas:
Alright, good to know about `Interval` being public! I will see when I can get to it for the PR, meanwhile to give you an idea for what I'm doing exactly is I am saving meta information of the node for the output tree so I can have reflection, here's the snippet copy paste but you will understand:

```js
function meta (src) {
  const _meta = { node: src }
  if (src.source) {
    // TODO: this is inefficient
    _meta.pos = src.source.getLineAndColumn()
    _meta.message = src.source.getLineAndColumnMessage()
  }
  return { _meta }
}
```

[...]

PR #285, opened 2021-01-13
feat: add `toString()` method in `getLineAndColumn()` result object

stagas:
This refactors the resulting object from `getLineAndColumn()` to cast to a message when using `toString()` on it. Other interfaces retain their signature but `getLineAndColumnMessage()` now uses this API and its previous implementation was extracted as a private util.


stagas:
Related #281

pdubroy:
Thanks!

PR #288, opened 2021-02-19
simple lisp example

szymonkaliski:
This PR introduces a simple lisp implementation.

The code is parsed to an AST, and then evaluated in a specified environment.
 
As a language this Lisp is probably pretty incomplete, but extending the core functions should be fairly straightforward.
This implementation hopes to sit between easy examples like math and csv, and more complex ones like ecmascript.

One thing that I'm unsure about is the implementation of macros in here, @pdubroy already mentioned that it might make sense to have them as a separate pass before "eval", which seems like a good idea - let me know if this is a blocker for getting this merged, and if not then I'd prefer to leave it as "nice to have".

### DIFF
--- a/examples/simple-lisp/simple-lisp.js
+++ b/examples/simple-lisp/simple-lisp.js
@@ -1,0 +1,411 @@
+const ohm = require("ohm-js");
+
+const grammar = ohm.grammar(`
+Lisp {
+  Program = Sexp*
+
+  Sexp = "(" Stmt ")"
+
+  Stmt = quasiquote (arg | argList ) -- quasiquote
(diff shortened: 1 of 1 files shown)


pdubroy:
Thanks for submitting this PR! Do you mind running `npm run lint -- --fix` in the repo root? That should fix up most of the lint issues.

szymonkaliski:
Sure yea - just pushed! I missed the CI issues :)

pdubroy:
Actually I'm going to go ahead and merge this, since that will make it easier for me to iterate on it a bit.

szymonkaliski:
awesome! let me know if there's anything I can do to help!

PR #308, opened 2021-08-02
BREAKING CHANGE: Don't use the default action if only child is an iteration node

pdubroy:
### Changes

Previously, the default action could be used for a non-terminal node if it had only a single child. This PR adds an additional check that the child is not an iteration node. **This is a breaking change.**

### Rationale

This ensures that the default action is type-safe. In TypeScript terms, if an operation has a return type `T` for non-terminal nodes, then by default it returns `T[]` for iteration nodes. The old behaviour meant that the default action could also return `T[]` for a non-terminal node, if its only child was an iteration node. E.g., take the following grammar and operation:

```
const g = ohm.grammar(`G {
  word = letters
  letters = letter+
}`);

const s = g.createSemantics().addOperation('toUpperCase()', {
  letter: l => l.sourceString.toUpperCase(),
})
```

Although the only action we've written returns `string`, the default action will be used for _word_ and _letters_, and return `string[]`:

```
> s(g.match('gg')).toUpperCase()
[ 'G', 'G' ]
```

In JavaScript, this behaviour can be surprising. For TypeScript (and also for JavaScript), it's desirable for all semantic actions to have the same return type, which wouldn't be possible under the current behaviour. We can't just make the return type of non-terminal actions `T | T[]`, because that would mean that the *_iter* action returns `(T | T[])[]`, which leads to infinite regress. By not using the pass-through action if the only child is an iteration node, we avoid this problem altogether.

### Implications

Some existing code that relies on the pass-through action may now trigger a "missing semantic action" error. The fix this, you can add a new action for the affected node that manually implements the pass-through behaviour. E.g., see [this action](https://github.com/harc/ohm/pull/308/files#diff-f2866739b3e63fb1a17d5b23decd6a3df91d68bb8f8f5a8380fe11da5a5e4f4fR262) which required modification in Ohm's examples.


pdubroy:
@alexwarth @yoshikiohshima @rtoal — would love some feedback on this. Let me know what you think!

pdubroy:
Related: https://github.com/harc/ohm/issues/309

pdubroy:
Gonna go ahead and merge this one. Comments after the fact are still welcome :-)

PR #337, opened 2021-11-14
Correct usage of "semantic" rule

harrisi:
This applies to syntactic rules, not "semantic" rules, and may lead to confusion.


pdubroy (approved):
Good catch, thanks! For some reason I keep making that typo.

PR #344, opened 2022-01-03
Make deferred init works for rollup

megabuz:
I've tried to use `ohm-js` with rollup and I've faced the issue of circular dependency in `pexprs.js`.

Based on specification all imports should be evaluated prior to the body of module doing the importing, so in case when `pexprs` requires bunch of extensions, e.g. `pexprs-allowsSkippingPrecedingSpace`, rollup includes these extensions in the final bundle before `pexprs`. However they depend on `pexprs` and I'm receiving error similar to https://github.com/harc/ohm/issues/185#issuecomment-364246068

In this PR I moved body of `pexprs` to separate file. It should make rollup module ordering work correctly.


megabuz:
I found several similar problems, I'm trying to fix them

megabuz:
So upon further testing I found that any `require` should be able to execute correctly before module that calls it. 
I moved some deferred inits from `main.js` and `Semantics.js` to separate files to break this order dependency

pdubroy:
@megabuz Thanks for taking this on! Is this PR ready for review? It seems to contain some unrelated changes. Can you trying rebasing your branch onto master?

megabuz:
@pdubroy it is ready to review. 
Could you point out unrelated changes? All changes should be related to breaking circular dependencies

pdubroy:
@megabuz Great! And sorry — my mistake, at a quick glance I thought some of these changes were related to something else.

Thanks again, and I'll try to review this in the next day or two.

pdubroy:
Awesome, thanks for doing this! It looked good to me so I've gone ahead and merged it.

pdubroy:
Just published v16.1.1 which contains these changes.

PR #350, opened 2022-01-08
feat: Add proper support for ES modules via a Rollup build step

### LINKED ISSUE #185: ES6+ module
It would be great if ES6+ modules were supported and user could simply import the modules in a standard way. Any plans to do that? Thanks

pdubroy:
Fixes #185. Now that Ohm can be built with Rollup (see #344), we can ship a proper ES module via a Rollup build step.


packages/ohm-js/src/errors.js
```diff
@@ -4,7 +4,7 @@
 // Imports
 // --------------------------------------------------------------------
 
-const pexprs = require('./pexprs');
+const pexprs = require('./pexprs-main');
```
pdubroy, line 7: Had to add this to avoid a warning about circular dependencies from Rollup.

pdubroy:
@megabuz: Please take a look and let me know if this looks like a reasonable approach to you.

pdubroy:
Published this as `ohm-js@16.2.0-pre.esm`.

pdubroy:
Going to merge this now since [it seems to fix the problem](https://github.com/harc/ohm/issues/185#issuecomment-1008004355). Happy to address anything else in a follow-up.

megabuz:
Thank you! I'm glad ohm receives better ESM support.
I'm going to test it in my project

PR #352, opened 2022-01-08
Option to generate esm module from @ohm-js/cli generateBundles

megabuz:


pdubroy (approved):
Awesome, thanks for this! Looks like the test snapshots just need updating; I'll do that and  merge manually.

PR #357, opened 2022-01-27
fix: Wrong parameters

kassadin:
The editor does not display the error message correctly

eg. 
```
Test{
  a = alnum
  b = a<x> 
}
```
editor need full err message with `interval`, line 34

https://github.com/harc/ohm-editor/blob/bb5f9a15d3d0991818139b0fd6e6b06356883751/src/index.js#L26-L42

but ohm hide it 

https://github.com/harc/ohm/blob/cf1e3abfd61944cfa6784176dcf82910a90953b4/packages/ohm-js/src/errors.js#L129-L140

line 139 should be `expr`

sorry I don't know how to test, please check it

### DIFF
--- a/packages/ohm-js/src/errors.js
+++ b/packages/ohm-js/src/errors.js
@@ -136,7 +136,7 @@ function wrongNumberOfArguments(ruleName, expected, actual, expr) {
       ', got ' +
       actual +
       ')',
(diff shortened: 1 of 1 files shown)


pdubroy:
Thanks for the PR!

PR #451, opened 2023-05-09
Grammars for ES2015-ES2022

elgertam:
I have added preliminary grammars for all JavaScript releases from ES2015 through ES2022

- `generate-ecmascript-grammar.mjs` and `grammarkdown.ohm` have been updated to parse later ES specs
- `extract-grammarkdown.mjs` and `dedent.py` have been included to create a clean Grammarkdown grammar from an ES spec
- Extracted & manually cleaned ES grammars for each release
- Per-grammar overrides have been moved from `generate-ecmascript-grammar.mjs` to an `override.json` file, along with productions containing reserved words
- Grammarkdown grammars & overrides have been organized into folders for each version of ES
- Generated Ohm grammars for each ES version
  - Grammars appear to parse JS programs correctly in the Ohm editor
  - Using the esm module loader syntax is only supported if the start symbol is changed to `Module`


examples/ecmascript/package.json
```diff
@@ -40,5 +40,10 @@
     "src",
     "!**/*.test.*",
     "!**/testdata/**"
-  ]
+  ],
+  "dependencies": {
+    "@pkgjs/parseargs": "^0.11.0",
+    "ecmarkup": "^17.0.0",
```
pdubroy, line 46: Same, is this used/required?

pdubroy (commented):
Thanks so much for this work! I'm excited that we're so close to having some modern ECMAScript grammars.

So far I've only had time to have a cursory look through it, but I'm hoping I'll have a chance to play with it tomorrow and then can likely optimistically merge it.

examples/ecmascript/package.json
```diff
@@ -40,5 +40,10 @@
     "src",
     "!**/*.test.*",
     "!**/testdata/**"
-  ]
+  ],
+  "dependencies": {
+    "@pkgjs/parseargs": "^0.11.0",
```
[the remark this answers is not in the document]
elgertam: `parseArgs()` is in the Node standard lib in versions 18+. This one is a polyfill for older versions.

examples/ecmascript/package.json
elgertam: This one is detritus. I'll remove & repush :)

pdubroy:
I've merged into a branch for now, so we can both collaborate on that: https://github.com/ohmjs/ohm/tree/elgertam-ecmascript

PR #457, opened 2023-08-07
Add Grammar Include Support

### LINKED ISSUE #286: Is there's a way to include grammar definitions from other file
Correct me if I'm wrong, but from what I see  there's no way to include grammar definitions from one file into another one.

My questions are following:

1. Is it indeed the case?
2. If yes, can we discuss introduction of such feature? I'm OK with participating if needed.

LiamRiddell:
Fixes #286 

I've implemented a fix that prevents VS Code complaining about LF vs CLRF line endings. This uses the solution provided in [this article](https://warlord0blog.wordpress.com/2019/09/04/vscode-crlf-vs-lf-battle/) and enforces LF line endings.

# Implementation (2.0)
- [X] Revert the original commit for `ohm-cli`
- [X] Add a new grammar rule for `include '...'` in the [ohm-js grammar]( https://github.com/ohmjs/ohm/blob/main/packages/ohm-js/src/ohm-grammar.ohm)
- [x] Introduce unit tests for the new `include` statement in the grammar
- [x] Implement visitor for `Document`, `Includes`, and `Include` rules
- [x] Implement optional `options` object for the `ohm.grammar` function in `ohm-js`
- [x] Add new option called `fetchImport` that is provided via optional `options` object (Add circuit breaker for preventing infinite loop includes)
- [ ] Add default implementation for Node and Browser, utilising `.node.js` and `.browser.js` file type support (optional)

# Thoughts
- [ ] How does this work when bundling with CLI for TypeScript? I feel we should have two modes, resolve and passthrough mode.


pdubroy:
Hi Liam,

Thanks for the putting in the work to prototype this! Adding support for imports is something we've thought about over the years, but have hesitated to implement. Mainly, my concern is — what happens if someone is trying to load the grammar via `ohm.grammar(...)` rather than using the CLI? And then, the question is how to handle imports when Ohm is running in Node, vs. the browser?

This is a feature that I'd be happy to add, but if we're going to do it, I'd like to make sure we think it through and implement it fully. If it's something you're up for, we could collaborate on it.

I'll leave some notes on #286 about some of the things I'd like to address in a complete implementation.

pdubroy:
Whoops, didn't mean to close this PR. I think we keep it open and iterate on it.

LiamRiddell:
> Hi Liam,
> 
> Thanks for the putting in the work to prototype this! Adding support for imports is something we've thought about over the years, but have hesitated to implement. Mainly, my concern is — what happens if someone is trying to load the grammar via `ohm.grammar(...)` rather than using the CLI? And then, the question is how to handle imports when Ohm is running in Node, vs. the browser?
> 
> This is a feature that I'd be happy to add, but if we're going to do it, I'd like to make sure we think it through and implement it fully. If it's something you're up for, we could collaborate on it.
> 
> I'll leave some notes on #286 about some of the things I'd like to address in a complete implementation.

Hi Patrick,

I'd love to collaborate, this as you mentioned was just a fix to get it working for my use case but I was keen to bring it in language as opposed to a CLI-only option.

```ts
ohm.grammar(`
  include './foo.ohm'

  G {
    ...
  }
`, {
  fetchGrammar(relativeUrl) {
    return fs.readFileSync(relativeUrl, import.meta.url, 'utf-8');
  }
})
```

This is a good starting point. Which allows the user to provide the method of retrieval. However, I'd expect to see:

* Node - Read the file from disk based on the relative. 
* Browser - Fetched from url based on the relative path.

I'll add a new grammar rule for `include '...'` in the ohm-js grammar:
https://github.com/ohmjs/ohm/blob/main/packages/ohm-js/src/ohm-grammar.ohm.

Furthermore, I'll get some tests in for that too.

pdubroy:
A couple of the tests depend on the generated bundles, so I think you need to run `pnpm build` before `pnpm test`.

(It's not ideal — maybe we should run it `build` automatically before the `test`, but that can result in wasted work in a lot of circumstances.)

packages/ohm-js/src/ohm-grammar.ohm
```diff
@@ -61,6 +70,12 @@ Ohm {
     | terminal                                       -- terminal
     | "(" Alt ")"                                    -- paren
 
+  include
+  	= caseInsensitive<"include">
```
pdubroy, line 74: I don't think this should be case insensitive; it's not very common for languages to have case-insensitive keywords.

LiamRiddell:
> A couple of the tests depend on the generated bundles, so I think you need to run `pnpm build` before `pnpm test`.
> 
> (It's not ideal — maybe we should run it `build` automatically before the `test`, but that can result in wasted work in a lot of circumstances.)

That makes sense, I just ran a `pnpm build` followed by `pnpm build` and it's working as expected with all but two tests passing.

1. tracing › memoization
2. tracing › bindings

Do these usually fail?

pdubroy:
> Do these usually fail?

Yes, those are marked as known failures. The output is slightly confusing but ava allows you to mark tests as "known failures" so that `pnpm test` succeeds (= return code 0) but still reports those as known failures.

LiamRiddell:
> > Do these usually fail?
> 
> Yes, those are marked as known failures. The output is slightly confusing but ava allows you to mark tests as "known failures" so that `pnpm test` succeeds (= return code 0) but still reports those as known failures.

Interesting, thanks for the insight. That makes sense to why I'm getting the return code 0.

LiamRiddell:
@pdubroy

I've made good progress, the import tests are working as expected. However, I'm getting the same two tests failing.

ohm-syntax > bootstrap it can produce a grammar that works
```
  test/test-ohm-syntax.js:1393

   1392:   test('it can produce a grammar that works…
   1393:     const g = buildGrammar(ns.Ohm.match(ohm…
   1394:     assertSucceeds(                         

  Error thrown in test:

  Error {
    message: 'Cannot use a MatchResult from grammar \'Ohm\' with a semantics for \'Ohm\'',
  }

  › ASemantics (file://src/Semantics.js:524:13)
  › buildGrammar (file://src/buildGrammar.js:301:10)
  › file://test/test-ohm-syntax.js:1393:15
```
ohm-syntax > bootstrap full bootstrap
```
test/test-ohm-syntax.js:1445

   1444:   test('full bootstrap!', t => {            
   1445:     const g = buildGrammar(ns.Ohm.match(ohm…
   1446:     const gPrime = buildGrammar(g.match(ohm…

  Error thrown in test:

  Error {
    message: 'Cannot use a MatchResult from grammar \'Ohm\' with a semantics for \'Ohm\'',
  }

  › ASemantics (file://src/Semantics.js:524:13)
  › buildGrammar (file://src/buildGrammar.js:301:10)
  › file://test/test-ohm-syntax.js:1445:15
```

Both are triggering the MatchResult error, when debugging it looks like because the grammar source is different:

[const cst = matchResult._cst;](https://github.com/LiamRiddell/ohm/blob/87732d4ffa38856bb66bb8c4860218481f02a1a2/packages/ohm-js/src/Semantics.js#L522-L531)

It looks like the grammar is not being updated? Which is confusing why the internal grammar is different. Am I missing something here?

image

# Current process
1. Make a change
2. `pnpm build`
3. `pnpm test`

LiamRiddell:
Okay, I've fixed the remaining tests 👍

packages/ohm-js/extras/extractExamples.js
```diff
@@ -73,6 +73,17 @@ const semantics = grammars.OhmWithExamples.createSemantics().addOperation('hasEx
 });
 
 semantics.addOperation('examples', {
+  Document(includesIter, grammarIter)
+  {
+    includesIter.examples();
```
pdubroy, line 78: I don't think it makes sense for `include` lines to have embedded examples. So you can change this as follows:

```suggestion
```

packages/ohm-js/extras/extractExamples.js
```diff
@@ -73,6 +73,17 @@ const semantics = grammars.OhmWithExamples.createSemantics().addOperation('hasEx
...
+  {
+    includesIter.examples();    
+    return grammarIter.examples();
+  },
+  Includes(includesIter) {
+    return includesIter.children.flatMap(c => c.examples());
+  },
+  Include(_, _la, relativePathNode, _ra) {
+    return null;
+  },
```
pdubroy, line 86: And then you should be able to remove these actions:

```suggestion
```

packages/ohm-js/src/ohm-grammar.ohm
```diff
@@ -61,6 +70,12 @@ Ohm {
     | terminal                                       -- terminal
     | "(" Alt ")"                                    -- paren
 
+  include
+  	= "include"
+    
+  relativeFilePath
```
pdubroy, line 76: What's the reason for choosing to allow these specific characters? Maybe I'm wrong, but it seems like it may be overly restrictive.

An alternative would be for Ohm to be completely agnostic about paths, and accept any character inside the quotes.

packages/ohm-js/src/buildGrammar.js
```diff
@@ -19,16 +20,79 @@ function namespaceHas(ns, name) {
...
   // A visitor that produces a Grammar instance from the CST.
   const helpers = metaGrammar.createSemantics().addOperation('visit', {
+    Document(includesNode, grammarsNode)
+    {
+      const resolvedIncludes = includesNode.visit();
+
+      // We need to rebuild the grammar match with the resolved includes substituted.
+      // Note: It's important we prevent any deeper visits in this tree as it's now pointless.
+      if (resolvedIncludes.length > 0) {
+        rematchInput(resolvedIncludes);
```
pdubroy, line 75: This doesn't seem like the right way to handle this. Naively I'd expect a call to `compileAndLoad` for each include.

And modifying the original source and calling `match` again shouldn't be necessary — we've already got `namespace` in which we look up other grammars. Can we visit includes, and call `compileAndLoad` on each of the includes to the namespace?

Which makes me realize that there is an unresolved design decision here, about whether includes are also "exported" or not (i.e., whether or not they are present in the final namespace that's handed back to the user). For simplicity I think it's fine if they are for now, but maybe there is an argument that they shouldn't be.

packages/ohm-js/test/test-ohm-syntax.js
```diff
@@ -1382,22 +1382,22 @@ describe('bootstrap', test => {
   const ns = ohm.grammars(ohmGrammarSource);
 
   test('it can recognize arithmetic grammar', t => {
-    assertSucceeds(t, ns.Ohm.match(arithmeticGrammarSource, 'Grammar'));
```
pdubroy, line 1385: Were these changes necessary, or is this an unrelated cleanup?

pdubroy (commented):
Looks like a great start! I have a few questions, and a request to change the implementation a bit.

packages/ohm-js/src/ohm-grammar.ohm
LiamRiddell: This is following security standards to always use an allowlist when you know the expected content format. If you want to accept the risk and open it to any character between the quoutes we can do that.

packages/ohm-js/test/test-ohm-syntax.js
LiamRiddell: Since we added new `Document` rule to grammar the default rule is now `Document`, I feel the tests should be validating against the new grammar and not a sub-rule (e.g. `Grammar`) of the main grammar. Hence the change and resulting fixes to array accesses below.

Issue #456, opened 2023-07-21
cli.js is reading incorrect package.json file for version


andysturrock:
Line 6 of cli.js reads the package.json file from the current directory:
`const {version} = JSON.parse(fs.readFileSync('./package.json'));`
If that package.json file does not have a version key then `npx ohm generateBundles` fails with:
`TypeError: Cannot read properties of undefined (reading 'option')`
If the local file does have a version then the -v flag will return that, rather than the version of ohmjs.

To get the current version of ohmjs, I think it should be:
`const {version} = JSON.parse(fs.readFileSync('./node_modules/@ohm-js/cli/package.json'));`

If you agree I am happy to submit a PR.

pdubroy:
Thanks for the bug report! I think a more appropriate fix would be to use a module-relative URL...e.g. something like:

```
fs.readFileSync(new URL('../package.json', import.meta.url));
```

andysturrock:
Yep that makes more sense.

PR https://github.com/ohmjs/ohm/pull/458 created

pdubroy:
This issue has been fixed.

PR #458, opened 2023-08-11
Fix package.json location (#456)

andysturrock:
From https://github.com/ohmjs/ohm/issues/456

Parse ohm's package.json file to obtain the version. cli.js was parsing the package.json in the current directory to obtain the package version.
It now parses its own package.json file, not the one from the project that is using ohm.


pdubroy:
Thank you for the fix!

PR #477, opened 2024-04-28
Es grammars fix

CrazyWinner:
Es2016 to Es2022 grammar wasn't parsing basic statements like "i++;". I was able to fix 2016 to 2021 but 2022 is stil broken.


pdubroy:
Just getting around to clearing a backlog of PRs. Thanks for contributing this! Any idea why 2022 is broken?

PR #478, opened 2024-04-29
Patch syntax-reference.md

ryangoree:
The non-escaped chars were causing the last lines to get rendered together:

image


pdubroy:
Thanks!

PR #483, opened 2024-06-19
Fix typo in patterns-and-pitfalls.md

ThatBlockyPenguin:
Line 99: change "right-assiciative" to "right-associative".


pdubroy:
Thanks!

PR #512, opened 2025-06-10
Adds all unicode categories to builtin rules

kgwinnup:


kgwinnup:
`startChar = "_" | unicodeChar<"Lu"> | unicodeChar<"Ll"> | unicodeChar<"Lt"> | unicodeChar<"Lm"> | unicodeChar<"Lo">`

I like it, this is an excellent approach. I'll wait for you as you obviously have more experience here than I.

Cheers

PR #528, opened 2025-09-17
Improve Instantiating Grammars documentation with verbose examples

karloscarweber:
Hi Friends.

This PR improves the documentation around Instantiating Grammars somewhat by providing verbose examples. We could simplify the examples to save space, and add clarity, but I think verbose examples are usually helpful.

The desire to add this came about because I was a bit confused about how to actually instantiate a grammar that inherits from another. The documentation is clear, but I had to read it thoroughly a couple of times, and ask for some help, before I understood what was going on.


karloscarweber:
Tests are failing. It's stemming from the code examples. I'll add some markscript checks and run through the test suite locally to ensure that it will pass. Then upload my changes.

